Issue #3, opened 2020-11-26
Next Steps


Sean-Der:
@masterada @at-wat Just starting a tracking issue, feel free to add anything to this list you think are important! I will also be helping after `v3.0.0` is tagged :)

* [x] Create examples
* [x] Move NACK to dedicated pkg. Should top-level just be structure, and each specific topic be their own folder?
* [x] Tests
* [x] Finish README
  - [x] Elevator Pitch
  - [x] Roadmap/Interceptor Types
  - [x] Motivations for Interceptors/why we did it this way
  - [x] Architecture/Design Document (in own .md file?)

Sean-Der:
@aler9 also would love your opinion :)

masterada:
Do we need WASM build in this project? If not, I think we should remove WASM test.

Sean-Der:
@masterada I think WASM tests might actually work with no modifications!

`pion/webrtc` isn't very WASM friendly. However `pion/rtp` and `pion/rtcp` should be just fine. I think it is worth having it work. In the future WASM support could unlock some really interesting stuff.

If we can get socket support in the browser we could run Pion in the browser right away! I don't have any concrete plans, but I think it could be something that will suprise us eventually :)

Sean-Der:
What do you think of moving the tests from `pion/webrtc` into here for interceptors? Just trying to find a good mix between bringing the time down to run all the tests without lowering the software quality :)

For me `pion/webrtc` sits at ~90 seconds, which isn't terrible but we can always do better :)

masterada:
> What happens if you just remove all exclusions?
Yes, they run, just seemed a bit unnecessary to run tests twice, but doesn't really hurt I guess.

> What do you think of moving the tests from pion/webrtc into here for interceptors?
The test in interceptor_test is there to make sure interceptors are properly called everywhere (so it's more of a test for the intergration of interceptors into webrtc), so I think we should leave it there. Tests for specific interceptors can go into this packet (eg: see my nack pr)

PR #4, opened 2020-11-27
Added Nack interceptors

masterada:
Ref: https://github.com/pion/interceptor/issues/3


.gitignore
```diff
@@ -0,0 +1,24 @@
+### JetBrains IDE ###
```
Sean-Der, line 1: :+1: on this! Would you mind adding it to github.com/pion/.goassets 

We use that to sync common files across all repos

nack/receive_log.go
```diff
@@ -0,0 +1,141 @@
...
+	"strconv"
+	"sync"
+)
+
+var (
+	allowedReceiveLogSizes     map[uint16]bool
+	invalidReceiveLogSizeError string
+)
+
+func init() {
```
Sean-Der, line 14: I would generate `allowedReceiveLogSizes` in `NewReceiveLog` I don't think that is going to be a hot spot.

With `invalidReceiveLogSizeError` I would do a static error, and then include the metadata. That way people can do `errors.Is`

```
InvalidReceiveLogSizeError = errors.New("Invalid receive log size")

...
...
return fmt.Errorf("%w: %d is not a valid size", InvalidReceiveLogSizeError, requestedSize)
```

nack/receiver_interceptor.go
```diff
@@ -0,0 +1,175 @@
...
+		receiveLogs: &sync.Map{},
+		close:       make(chan struct{}),
+		log:         log,
+	}, nil
+}
+
+// BindRTCPWriter lets you modify any outgoing RTCP packets. It is called once per PeerConnection. The returned method
+// will be called once per packet batch.
+func (n *ReceiverInterceptor) BindRTCPWriter(writer interceptor.RTCPWriter) interceptor.RTCPWriter {
+	n.m.Lock()
```
Sean-Der, line 48: `defer n.m.Unlock()` should be ok here!

test/stream.go
```diff
@@ -0,0 +1,148 @@
+package test
```
Sean-Der, line 1: Is this a helper package for writing tests with `interceptor` ? Would `internal/test` make more sense maybe?

Sean-Der:
@masterada this is really fantastic! I wish I had this 5 years ago when learning. The code is so clear, I can't wait for people to go read this and understand RTC in a week instead of years :)

Sean-Der:
What do you think of doing a functional pattern for config, then we can add/remove options as we please. @at-wat did this in `pion/srtp` https://github.com/pion/srtp/blob/master/context.go#L61

In the future I think we can expand things a lot more!
* Distinct stream for NACKs
* Do dynamic instead of static etc...

Or we can have distinict objects for the different types.

nack/send_buffer.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	uint16SizeHalf = 1 << 15
+)
+
+var (
+	allowedSendBufferSizes     map[uint16]bool
+	invalidSendBufferSizeError string
+)
+
+func init() {
```
Sean-Der, line 19: Same comments as above re: init/dynamic errors.

.gitignore
masterada: Will do that on the weekend when I have some time, along with all other comments. Thanks for the review, these are really useful comments :)

Sean-Der:
@masterada rebased off of master, reviewing again!

nack/generator_interceptor.go
```diff
@@ -0,0 +1,181 @@
...
+
+	return r, nil
+}
+
+// BindRTCPWriter lets you modify any outgoing RTCP packets. It is called once per PeerConnection. The returned method
+// will be called once per packet batch.
+func (n *GeneratorInterceptor) BindRTCPWriter(writer interceptor.RTCPWriter) interceptor.RTCPWriter {
+	n.m.Lock()
+	defer n.m.Unlock()
+	select {
```
Sean-Der, line 56: Mind adding a `n.isClosed()` I like this pattern, but newer Go devs have gotten confused/asked me a lot about it. Just helps to make a little more self documenting!

Sean-Der:
@masterada this is really great, approved! I especially love how readable the tests are. I think the Public API is rock solid, and the internals are good. Lots of WaitGroup, Mutex and Channels make for challenging code, but that is the problem space!

I think if we have this + Receiver/Sender reports and enable them by default that would make a perfect `v3.0.0`

@at-wat would love if you could review/see if this captures the spirit? This will be a blueprint for lots of future work, so would love to have it.

internal/test/stream.go
```diff
@@ -0,0 +1,161 @@
+package test
```
at-wat, line 1: Something like `mockstream` may be better package naming.

nack/generator_interceptor.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/pion/rtp"
+)
+
+// GeneratorInterceptor interceptor generates nack feedback messages.
+type GeneratorInterceptor struct {
+	interceptor.NoOp
+	size        uint16
+	skipLastN   uint16
+	interval    time.Duration
+	receiveLogs *sync.Map
```
at-wat, line 20: Official document says: https://godoc.org/sync#Map
> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content. 

I think it's better to use strict type for this.

nack/receive_log_test.go
```diff
@@ -0,0 +1,134 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestReceivedBuffer(t *testing.T) {
+	for _, start := range []uint16{0, 1, 127, 128, 129, 511, 512, 513, 32767, 32768, 32769, 65407, 65408, 65409, 65534, 65535} {
+		start := start
+
```
at-wat, line 11: Looks better to make this block subtest.
```suggestion
		t.Run(fmt.Sprintf("StartFrom%d", start). func (t *testing.T) {
```

nack/responder_interceptor.go
```diff
@@ -0,0 +1,127 @@
...
+
+type localStream struct {
+	sendBuffer *sendBuffer
+	rtpWriter  interceptor.RTPWriter
+}
+
+// NewResponderInterceptor returns a new GeneratorInterceptor interceptor
+func NewResponderInterceptor(opts ...ResponderOption) (*ResponderInterceptor, error) {
+	r := &ResponderInterceptor{
+		NoOp:    interceptor.NoOp{},
```
at-wat, line 28: As it's not a pointer type, `NoOp` is automatically initialized.

nack/responder_interceptor.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		_, err := stream.rtpWriter.Write(p, interceptor.Attributes{})
+		if err != nil {
+			n.log.Warnf("failed resending nacked packet: %+v", err)
+		}
+	}
+}
+
+func nackParsToSequenceNumbers(pairs []rtcp.NackPair) []uint16 {
```
at-wat, line 120:
```suggestion
func nackPairsToSequenceNumbers(pairs []rtcp.NackPair) []uint16 {
```

nack/responder_option.go
```diff
@@ -0,0 +1,21 @@
+package nack
+
+import "github.com/pion/logging"
+
+// ResponderOption can be used to configure ResponderInterceptor
+type ResponderOption func(s *ResponderInterceptor)
```
at-wat, line 6: I think it's better to make it
```suggestion
type ResponderOption func(s *ResponderInterceptor) error
```
and validate the arguments. (`ResponderSize` should be 2^n)

Sean-Der:
@masterada If you don't have the availability I am happy to pull in the suggested changes :)

No rush, I just don't want to overload you! What do you think of having `pion/webrtc` go out with Receiver Reports + NACK by default? You think good/bad idea.

Or should we push `v3.0.0` with no interceptors, and do a minor release with them?

nack/generator_interceptor.go
masterada: It also states: 

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times

which is the exact use case here

nack/responder_option.go
masterada: I think it would be better to do this only for the size option, eg:

// ResponderSize sets the size of the interceptor.
// Size must be one of: 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768
func ResponderSize(size uint16) (ResponderOption, error) {
	
	_, err := newSendBuffer(r.size)
	if err != nil {
		return nil, err
	}

	return func(r *ResponderInterceptor) {
		r.size = size
	}, nil
}

This way we can also remove the error return from NewResponderInterceptor
same applies to GeneratorInterceptor

masterada:
@Sean-Der sorry, I just started work at a new place and it comsumes more energy than I expected. Not sure when I could finish these comments, probably only next week (my weekend is kinda budy too). Please if you have the time feel free to finish this up. Also there is a TODO to move sequence numbers -> nack pair conversion to rtcp packet, because the other way is alread there. It's not really important, can be done sometime later too.

I replied to 2 comments @at-wat , other than that I agree with you in everything.

> What do you think of having pion/webrtc go out with Receiver Reports + NACK by default? You think good/bad idea.

I think it would be nice to add at least 1 interceptor to 3.0.0, so we can be sure the interceptor integration works, and also create the RegisterDefaultInterceptors method that actually does something. RR/SR and TCC would be nice too but they can come at 3.1 too or something.

Also note that in order for nack.GeneratorInterceptor to work, you need to call s.SetSRTPReplayProtectionWindow(AT_LEAST_NACK_SIZE), see: https://github.com/pion/webrtc/pull/1549/files#diff-6e2e56e295d82766f7fdce880ab65f299e92aede458043e232610c9fe9166e03R58

Maybe a higher default value for that could work? I think google uses 10000 for the replay protection window size, not sure though, their c code is very hard to follow.

nack/receive_log.go
```diff
@@ -0,0 +1,130 @@
...
+	lastConsecutive uint16
+}
+
+func NewReceiveLog(size uint16) (*ReceiveLog, error) {
+	if !allowedReceiveLogSizes[size] {
+		return nil, errors.New(invalidReceiveLogSizeError)
+	}
+
+	return &ReceiveLog{
+		packets: make([]uint64, size/64),
```
jech, line 33: This should be an interface: if my application is already logging packets for some other reason, then I should be able to pass my receive log to the interceptor rather than keeping yet another copy of the log.

nack/receiver_interceptor.go
```diff
@@ -153,6 +153,7 @@ func (n *ReceiverInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 }
 
 func nackPairs(seqNums []uint16) []rtcp.NackPair {
+	// TODO: I think this shoud be moved to rtcp package
```
jech: Agreed.  The function I'm using for that in the Unnamed SFU has the following shape:
```
func ToBitmap(seqnos []uint16) (first uint16, bitmap uint16, remain []uint16)
```
it returns the first NackPair (as first/bitmap) and anything remaining afterwords.  You're supposed to loop until `remain` is empty.

nack/receiver_interceptor.go
```diff
@@ -0,0 +1,174 @@
...
+
+// BindRemoteStream lets you modify any incoming RTP packets. It is called once for per RemoteStream. The returned method
+// will be called once per rtp packet.
+func (n *ReceiverInterceptor) BindRemoteStream(info *interceptor.StreamInfo, reader interceptor.RTPReader) interceptor.RTPReader {
+	hasNack := false
+	for _, fb := range info.RTCPFeedback {
+		if fb.Type == "nack" && fb.Parameter == "" {
+			hasNack = true
+		}
+	}
```
jech: There should be a utility function for that.

nack/receiver_interceptor.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return nil
+}
+
+func (n *ReceiverInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
+	defer n.wg.Done()
+
+	senderSSRC := rand.Uint32()
+
+	ticker := time.NewTicker(n.interval)
```
jech: Rather than doing this at set intervals, wouldn't it be better to trigger when packet n is missing but packet n + k has been received, for some suitable value of k?  In the Unnamed SFU, k is computes as: at least 3 (since that is what TCP assumes, and anything that makes stronger assumptions than TCP is likely to break in the Internet), and no more than the tolerable latency divided by the packet rate.

nack/receive_log.go
masterada: Buffers should not be shared. Here is my reasoning:

receive_log:
Receive log stores only the fact that the packet arrived. That means it uses 1 bit per packet, for 8192 pacekts, it uses 1024bytes. If I remember correctly the mtu is about 1.5 kB, so that means the whole receive log is smaller than a single video packet.

send_buffer:
This is more tricky, because it stores every packet. If you have multiple outgoing stream, like in the case of a SFU, it will store the packets for every stream. But:
- packet payload is a slice, meaning the underlying data is not copied, it's only stored once
- header will be mostly different for every sfu client, because payload type, extensions, ssrc, etc. can differ
If reusing a buffer, you would have to fix all headers every time you send out a packet. This might work for an sfu server, but there are more complicated use cases than that.
- in a centralized buffer it's much harder to configure when a packet can be discarded (eg: in the future the nack sender interceptor can listen to tcc packets as well, allowing it to drop packets based on it)

The final reason I'm against a single shared buffer is jitter handling. You might want to use the same stream with and without a jitter buffer at the same time. We have an aplication that saves a stream to a webm file on disk, but also forwards the stream to a single viewer. We then proxy the tcc packets from the viewer back to the broadcaster. In this case, the file save part needs a jitter buffer, but the proxy part does not, because the jitter buffer messes with the packet delays from the viewer's perspective, and messes up the whole tcc.

nack/receiver_interceptor.go
masterada: I think this can be improved in a later mr as this is an internal behaviour. Also there should be a limit to how many times a nack for a packet is requested. k is already configurable, it's the skipLastN parameter.

Sean-Der:
@masterada congrats on the new job! No worries at all, you have done some amazing work here. I am more then happy to finish out the rest :)

nack/receive_log.go
jech: Sorry, yes, I meant the send buffer.

In this comment, I'm not arguing against or in favour of a shared buffer.  I'm merely arguing that the send buffer should be an interface that the user implements.  Your current implementation is probably fine as the default, but the user should be able to provide a different implementation for the buffer.

nack/generator_interceptor.go
Sean-Der: Fixed!

internal/test/stream.go
Sean-Der: @at-wat what do you think of the name now? Feel free to edit the branch directly!

nack/responder_interceptor.go
Sean-Der: Fixed!

nack/responder_option.go
Sean-Der: Fixed! Both options now return errors. That could come in handy in the future also!

nack/receive_log_test.go
Sean-Der: Fixed!

nack/responder_interceptor.go
Sean-Der: Removed function entirely, used similar thing in `pion/rtcp`

nack/receiver_interceptor.go
Sean-Der: Moved! `NackPairsFromSequenceNumbers` is now in `pion/rtcp`

nack/receiver_interceptor.go
Sean-Der: Fixed! `streamSupportNack` in `pkg/interceptor`

nack/receiver_interceptor.go
Sean-Der: Created ticket!

I bet there are a lot of things we can do to make this better. I am really happy with what we have today though :)

Sean-Der:
@masterada @at-wat mind giving it another look? I went through and address everything in your reviews.

thanks!

masterada:
LGTM, but i think this should wait until the raw read and attributes changes are in.

Sean-Der (approved):

PR #12, opened 2020-12-07
Add read raw rtp packets

OrlandoCo:
#### Description

Adds the possibility to read a raw packet, this method will help to reuse byte slices and have more control over packet lifetime.

### DIFF
--- a/chain.go
+++ b/chain.go
@@ -59,6 +59,16 @@ func (i *Chain) BindRemoteStream(ctx *StreamInfo, reader RTPReader) RTPReader {
 	return reader
 }
 
+// BindRemoteRawStream lets you modify any incoming RTP raw packets. It is called once for per RemoteStream. The returned method
+// will be called once per rtp packet.
+func (i *Chain) BindRemoteRawStream(ctx *StreamInfo, reader RTPRawReader) RTPRawReader {
+	for _, interceptor := range i.interceptors {
+		reader = interceptor.BindRemoteRawStream(ctx, reader)
+	}
+
+	return reader
+}
+
 // UnbindRemoteStream is called when the Stream is removed. It can be used to clean up any data related to that track.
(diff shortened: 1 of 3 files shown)


Sean-Der:
How much of an improvement benchmark was this for `ion`? Is the extra copying really bad for packets :( 

Should the interceptor chain convert things as they pass through. What do you think should happen if you have three interceptors and they look like `Packet -> Raw -> Packet` ?

Sean-Der:
So I just realized is that Interceptors aren't executed at all for `Read` in pion/webrtc! So this isn't a performance issue, but a blocker on usability. 

I think we need to pass around `[]byte` or happy to hear alternatives. I am calling it a night, but this is for sure a `v3.0.0` blocker.

Sean-Der:
@at-wat @masterada if you have any ideas would love to hear them also :)  We are painfully close

masterada:
I left the simple Read out intentionally, because of the complexity they introduce:
A typical inteceptor will check the packet ssrc, read some extension info, maybe even peek into the payload. If there is a simple Read method it would mean that interceptors either:
a. Implement it, which means partially parsing the header in every interceptor, which might mean a worse performance implication than parsing the packet once. It also introduces a lot of complexity to the interceptor. 
b. Choose NOT to implement it, which would be more confusing than not supporting it in the first place.

In case of write it's even worse, because write interceptors may add extensions, which changes the header size, so the payload offset changes, so the packet payload must be copied into a new slice after the longer header.

My idea waa to leave the raw Read/Write for advanced use cases, but in those cases interceptors would not run. I tried to reflect this in the webrtc function docs, and obviously failed.

Can you describe the use case for this, and the reason raw Read is needed intead of ReadRTP? (if you just link the related code that would he helpful too).

Maybe we can do some softer parse before the interceptor chain that has better performance? Or some kind of alternate Packet type thats a slice, but with helper methods for header fields? I will keep thinking.

PR #18, opened 2020-12-12
Update RTP/RTCP Read to be []byte

### LINKED ISSUE #14: Interceptor only provide rtp.Packet based API, some users want to deal with []byte
@OrlandoCo just discovered this issue in [pion/webrtc](https://github.com/pion/webrtc/pull/1582) currently if a user does `Read` (not `ReadRTP`) it will skip the interceptors entirely.

Currently Interceptors deal entirely in rtp.Packet so `pion/webrtc` so we just don't use them in this case. We have the following choices. We have three different things we could do.

### Provide the ability to write `rtp.Packet`, `rtcp.Packet` and `[]byte`
Each Interceptor would provide the ability to do both. 

### Move to a `[]byte` based API
Instead of dealing with `rtp.Packet` in the public API everything will be `[]byte` based.

### Keep the current API
No changes

-----
These are my random thoughts as of today.

* If we switch to `[]byte` Interceptors will have to parse multiple times
  - Is this actually a problem, how expensive is parsing?
* If we switch to `[]byte` we can drop the dependency on `pion/rtcp` and `pion/rtp`
 - Do people care about this?
* Having multiple paths means more bugs/code to maintain

In pion/webrtc in the short term I propose switching `Read` to be fed via `Unmarshal` from `ReadRTP`. The extra operations are really unfortunate, but it unblocks the issue today.

Sean-Der:
Returning a rtp.Packet requires an extra allocation.

Resolves #14

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 
 * [Adam Kiss](https://github.com/masterada) - *Original Author*
 * [Sean DuBois](https://github.com/sean-der) - *Original Author*
+* [Atsushi Watanabe](https://github.com/at-wat)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/interceptor.go
+++ b/interceptor.go
@@ -43,13 +43,13 @@ type Interceptor interface {
 // RTPWriter is used by Interceptor.BindLocalStream.
 type RTPWriter interface {
 	// Write a rtp packet
-	Write(p *rtp.Packet, attributes Attributes) (int, error)
+	Write(header *rtp.Header, payload []byte, attributes Attributes) (int, error)
 }
 
 // RTPReader is used by Interceptor.BindRemoteStream.
 type RTPReader interface {
 	// Read a rtp packet
-	Read() (*rtp.Packet, Attributes, error)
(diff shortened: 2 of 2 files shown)


Sean-Der:
@masterada @OrlandoCo @tarrencev If I can get an approval on this I am gonna to update `nack` and then merge everything!

OrlandoCo (approved):
lgtm!

OrlandoCo:
@Sean-Der Should we add an special method for rtcp simulcast read or add option to pass the rid?

Sean-Der:
Hey @OrlandoCo 

I was able to make this work without making any interceptor changes! We know the SSRC when creating the interceptor so it works with no changes :) 

I am going to tag this and open my PR in `pion/webrtc` and with your sign-off we should be good :fire:

OrlandoCo:
> Hey @OrlandoCo
> 
> I was able to make this work without making any interceptor changes! We know the SSRC when creating the interceptor so it works with no changes :)
> 
> I am going to tag this and open my PR in `pion/webrtc` and with your sign-off we should be good

@Sean-Der But pion `ReadSimulcastRTCP` does not call interceptors, or it does on recent change?

Sean-Der:
It does on recent change. Everything *should* run through and interceptor now

Sean-Der:
If I missed something was an accident

PR #29, opened 2021-03-19
Fix time.Ticker leak

at-wat:
Ticker must be Stop()-ed after use.


Sean-Der (approved):

Sean-Der:
@at-wat if you are interested it would be worth aggressively refactoring these.

I have gotten reports from users that they see less packet loss when disabling them. My assumption is that all the alloc+copying in the NACK interceptor is an issue.

I also believe we don't evict the cache. We should do RTT*4 I think is the suggestion?

PR #38, opened 2021-04-22
Add SessionID to interceptors

### LINKED ISSUE #37: Interceptors cannot be shared between different sessions
Related to https://github.com/pion/webrtc/issues/1749

jeremija:
Closes #37. This commit addresses the changes discussed [here][1].

I added `SessionID` to all interceptor methods and ensured that the existing interceptors can be shared by different `PeerConnection`s.

I defined an `interceptor.Factory` following AtWat's suggestion. It took a while for this to click for me: I didn't realize that the `interceptor.Registry` contained already prebuilt interceptors. I like the idea of a `Factory` better than implementing `Copy`. I added the `interceptor.Registry.AddFactory` method and modified the internals a little. The new `interceptor.NewChainFromFactories` will now be called from the `Registry`.

The `interceptor.SharedFactory` allows us to easily keep the current behavior with existing interceptors.

I left a number of **TODO**s which are the main reason the build is currently failing. The main points are:

1. I think binding anything before calling `BindRTCPWriter` should be an error, because otherwise it becomes difficult to track state (e.g. if we've started a write loop). But I did not want to add an error return value to all methods before discussing it.

   If we do decide that calling `BindRTCPWriter` is the first step, then I think it would make sense to pass `RTCPWriter` as an argument to `interceptor.Registry.Build`?

2. We could modify the interceptors to only have a single `loop` goroutine which would go through all SSRCs for all bound sessions at once, but I'm not sure what's the best way to concurrently handle RTCP writes - I'm sure we wouldn't want a hanging RTCPWrite to slow down any other processing or writes.

   I also don't think that spinning up an unbounded number of goroutines just to do these writes is a good idea, so I'm not sure how to processed. I'm open to suggestions.

   Maybe this shouldn't be a part of this change, but it's something to think about.

[1]: https://github.com/pion/webrtc/issues/1749

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Sean DuBois](https://github.com/sean-der) - *Original Author*
 * [Atsushi Watanabe](https://github.com/at-wat)
 * [Alessandro Ros](https://github.com/aler9)
+* [Jerko Steiner](https://github.com/jeremija)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/chain.go
+++ b/chain.go
@@ -1,5 +1,7 @@
 package interceptor
 
+import "fmt"
+
 // Chain is an interceptor that runs all child interceptors in order.
 type Chain struct {
 	interceptors []Interceptor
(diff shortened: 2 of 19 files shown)


Sean-Der:
Fixed with fd0ded94670dacb67c583023b2c6ad073880a0fa

We instead went with a factory pattern. State can now be re-used inside of them.

PR #41, opened 2021-06-24
Naive PLI implementation

Antonito:
This PR implements a naive PLI Generator Interceptor.

The Naive PLI Generator sends a PLI packet for each new track that supports PLI, and then keep sending packets at a constant interval (defaults to 3sec).

The original motivation is to provide a easy way to have a very basic & inefficient but working implementation of PLI – and eventually replace PLI loops in `pion/webrtc/examples`:

```go
// Send a PLI on an interval so that the publisher is pushing a keyframe every rtcpPLIInterval
go func() {
	ticker := time.NewTicker(time.Second * 2)
	for range ticker.C {
		if rtcpErr := peerConnection.WriteRTCP([]rtcp.Packet{&rtcp.PictureLossIndication{MediaSSRC: uint32(track.SSRC())}}); rtcpErr != nil {
			fmt.Println(rtcpErr)
		}
	}
}()
```

### DIFF
--- a/pkg/pli/generator_interceptor.go
+++ b/pkg/pli/generator_interceptor.go
@@ -1,0 +1,146 @@
+package pli
+
+import (
+	"sync"
+	"time"
+
+	"github.com/pion/interceptor"
+	"github.com/pion/logging"
+	"github.com/pion/rtcp"
+)
+
+// GeneratorInterceptor interceptor sends PLI packets.
+// Implements PLI in a naive way: sends a PLI for each new track that support PLI, periodically.
+type GeneratorInterceptor struct {
+	interceptor.NoOp
+
+	interval           time.Duration
+	streams            sync.Map
+	immediatePLINeeded chan []uint32
+
+	log logging.LeveledLogger
+	m   sync.Mutex
+	wg  sync.WaitGroup
+
+	close chan struct{}
+}
+
+// NewGeneratorInterceptor returns a new GeneratorInterceptor interceptor.
+func NewGeneratorInterceptor(opts ...GeneratorOption) (*GeneratorInterceptor, error) {
+	r := &GeneratorInterceptor{
+		interval:           3 * time.Second,
+		log:                logging.NewDefaultLoggerFactory().NewLogger("pli_generator"),
+		immediatePLINeeded: make(chan []uint32, 1),
+		close:              make(chan struct{}),
+	}
+
+	for _, opt := range opts {
+		if err := opt(r); err != nil {
+			return nil, err
+		}
+	}
+
+	return r, nil
+}
+
+func (r *GeneratorInterceptor) isClosed() bool {
+	select {
+	case <-r.close:
+		return true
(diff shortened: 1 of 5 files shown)


pkg/pli/generator_interceptor.go
```diff
@@ -0,0 +1,146 @@
...
+	go r.loop(writer)
+
+	return writer
+}
+
+func (r *GeneratorInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
+	defer r.wg.Done()
+
+	ticker := time.NewTicker(r.interval)
+	defer ticker.Stop()
```
at-wat, line 89: Would be better to support disabling periodic PLI. (`r.interval == 0` to disable?)

pkg/pli/generator_interceptor.go
```diff
@@ -0,0 +1,146 @@
...
+// BindRemoteStream lets you modify any incoming RTP packets. It is called once for per RemoteStream. The returned method
+// will be called once per rtp packet.
+func (r *GeneratorInterceptor) BindRemoteStream(info *interceptor.StreamInfo, reader interceptor.RTPReader) interceptor.RTPReader {
+	if !streamSupportPli(info) {
+		return reader
+	}
+
+	r.streams.Store(info.SSRC, nil)
+	// New streams need to receive a PLI as soon as possible.
+	r.immediatePLINeeded <- []uint32{info.SSRC}
```
at-wat, line 132: It would be nice to have a method like
```go
func (r *GeneratorInterceptor) ForcePLI(ssrc uint32)
```
to trigger PLI from outside.

at-wat (commented):
I think periodic PLI is not needed by default since immediate PLI on BindRemoteStream is implemented.
Especially for low bitrate stream, adding more keyframe makes video quality lower.
What do you think?

Antonito:
> I think periodic PLI is not needed by default since immediate PLI on BindRemoteStream is implemented.
> Especially for low bitrate stream, adding more keyframe makes video quality lower.
> What do you think?

Yeah I agree, the goal of this Interceptor is not to provide a default PLI implementation but rather an optional "working" PLI implementation. (again, to simply examples and small PoC projects).
I'd love to have a default one, but it would need to parse RTP payloads 

I also agree with the two points you raised, I'll update my PR during the week end :)

pkg/pli/generator_interceptor.go
Antonito: This is now implemented [here](https://github.com/pion/interceptor/pull/41/files#diff-5f5c1c2e0b124989a33d5c38a83f0c09ecccfecd6277283325a914a12b620f4cR168) as 
```go
func (r *GeneratorInterceptor) ForcePLI(ssrc ...uint32)
```

Sean-Der:
Oh I really like this @Antonito!

After I merge this I am also going to update pion/webrtc to use this. It gives people an incentive/understanding of the value interceptors provide

PR #42, opened 2021-06-24
Packet Dumper Interceptors

Antonito:
This PR adds interceptors capable of dumping RTP & RTCP packets.

The interceptors directly write to a `io.Writer` which defaults to `os.Stdout` and can be configured by the user.

The RTP and RTCP packets can be dumped based on a filter provided by the user. By default, every packet is dumped.

Hopefully, this is what you had in mind @Sean-Der when writing the [2021 Goals](https://github.com/pion/webrtc/wiki/Goals-2021) 😁

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -6,6 +6,7 @@
 Adam Kiss <masterada@gmail.com>
 adamroach <adam@nostrum.com>
 aler9 <46489434+aler9@users.noreply.github.com>
+Antoine Baché <antoine@tenten.app>
 Atsushi Watanabe <atsushi.w@ieee.org>
 Jonathan Müller <jonathan@fotokite.com>
 Mathis Engelbart <mathis.engelbart@gmail.com>
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Check out the **[contributing wiki](https://github.com/pion/webrtc/wiki/Contribu
 * [Atsushi Watanabe](https://github.com/at-wat)
 * [Alessandro Ros](https://github.com/aler9)
 * [Mathis Engelbart](https://github.com/mengelbart)
+* [Antoine Baché](https://github.com/Antonito)
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/pkg/packetdump/filter.go
+++ b/pkg/packetdump/filter.go
@@ -1,0 +1,14 @@
+package packetdump
+
+import (
+	"github.com/pion/rtcp"
+	"github.com/pion/rtp"
+)
+
+// RTPFilterCallback can be used to filter RTP packets to dump.
+// The callback returns whether or not to print dump the packet's content.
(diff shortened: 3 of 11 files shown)


Antonito:
Concerning the thread-safety, I can either start a "writer's loop" in a goroutine and have a channel forward packet dumps to it, or simply add a mutex over the `Fprintf` calls.

I'm down for either way, the goroutine loop seems more efficient, but as these interceptors are destined for debug use only, maybe we'd be better off with the mutex and simpler code?

On a totally unrelated note, I'm quite concerned about pretty much every interceptor having to unmarshal packets in `BindRemoteStream` and `BindRTCPReader`. It seems every interceptor does this, so maybe there's a way to only do it once?

mengelbart:
Hi, I am using something similar to create RTP/RTCP packet dumps and it would be nice if I could replace it by this interceptor. I have a few thoughts about the implementation:

* Regarding repeated parsing in `BindRemoteStream` and `BindRTCPReader`: To me it seems, they mostly parse the whole packet, but then only look at the header. Repeated parsing of the header was also mentioned in [this discussion](https://github.com/pion/interceptor/issues/14) and the proposed solution was to parse it once and then store the header in the attributes. I don't think this has actually been implemented, but I guess we could just add another interceptor?
* I would like to use packet dumps to calculate metrics, so I would prefer a writer's loop in a goroutine instead of a mutex.
* For metrics it would also be nice to have the actual receive time dumped.
* Is it necessary to have two separate interceptors, or could it be combined in one interceptor, which dumps all incoming and outgoing RTP/RTCP packets?
* It would be nice if I could specify a custom format for packets, e.g. something like an option to set a `func(pkt *rtp.Packet) string` (maybe it should take another type, which only contains the packet and some other additional information, like receive time?).
* I would like to pass separate streams for RTP and RTCP, because it would make later analysis easier if I can just parse a CSV file for each stream.

Antonito:
Thanks for your feedback @mengelbart !

- Good point, I had missed this discussion. The proposed solution is something I'd be happy with – I don't think this has been implemented yet though (@Sean-Der can you confirm this?)
- Regarding the writer's loop and receive time, I agree – I only suggested the mutex workaround to quickly have a working interceptor that would fit most situations. If people already need/want to do precise metrics computation, yes, we need a dedicated go routine. The receive time can be one of those metrics.
  - However maybe a better design would be to add an additional Interceptor – which only captures additional data (such as the receive time) and have the PacketDumper interceptors able to print these information if present? That way users can easily combine interceptors to match their exact need – without performing unneeded work.
- It seems to me **exposing** two separate interceptors to the users is the way to go (for sake of ease & clarity of use), but we could implement them via a single 'internal' interceptor, to avoid code duplication.
  - However, as of today there's little code duplication and we might be better off with keeping things simple and clear?

Unfortunately I don't have time to work on this – and probably won't for a while. 
If you want to contribute to this PR, I'd be happy to review & discuss changes :)

PR #48, opened 2021-07-09
Update github.com/pion/rtp to v2

at-wat:
Bump major version.

### Reference issue
https://github.com/pion/rtp/issues/90


Sean-Der (approved):
Thank you so much for grabbing these!

PR #56, opened 2021-08-31
Fixing a comment

adamroach:
#### Description
Just clarifying an obvious copy/paste error

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -4,6 +4,7 @@
 # This file is auto generated, using git to list all individuals contributors.
 # see `.github/generate-authors.sh` for the scripting
 Adam Kiss <masterada@gmail.com>
+adamroach <adam@nostrum.com>
 aler9 <46489434+aler9@users.noreply.github.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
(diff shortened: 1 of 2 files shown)


Sean-Der:
:fire: really exciting to see you in this repo.

Anything we should change/improve? One big API change might be coming to accommodate https://github.com/pion/webrtc/issues/1749

Sean-Der:
@adamroach also mind if I add you to the Pion org? Just so you can create branches/pin against them etc... the less friction the more likely to keep you around :)

Sean-Der (approved):

adamroach:
> Anything we should change/improve? One big API change might be coming to accommodate pion/webrtc#1749

Thanks for the heads-up! I'm still feeling my way around how everything hangs together, so I don't have any clear suggestions for improvement yet. The examples in the project are really nice. I'm occasionally having difficulty figuring out what certain parameters mean, but the code is usually easy enough to read and figure them out.

> @adamroach also mind if I add you to the Pion org? Just so you can create branches/pin against them etc... the less friction the more likely to keep you around :)

Sure! I'm not sure how much time I'll have to contribute new functionality, but I believe in fixing things when I find issues, so I'll continue to do that. :)

PR #65, opened 2021-09-12
Use proper units with rtcp TWCC packets

### LINKED ISSUE #61: TWCC isn't performing as well as Browser <-> Browser
When using `save-to-disk` I am not seeing the BWE ramping up as fast as expected. 

When doing browser <-> browser I see a ramp up like this
Screen Shot 2021-09-08 at 2 29 03 PM

When doing browser <-> pion I see this
Screen Shot 2021-09-08 at 2 28 29 PM

Sean-Der:
pion/rtcp expects input to be in Nanoseconds, and does n / 250 during
the marshal.

Resolves #61

### DIFF
--- a/pkg/twcc/twcc.go
+++ b/pkg/twcc/twcc.go
@@ -51,28 +51,28 @@ func (r *Recorder) Record(mediaSSRC uint32, sequenceNumber uint16, arrivalTime i
 
 // BuildFeedbackPacket creates a new RTCP packet containing a TWCC feedback report.
 func (r *Recorder) BuildFeedbackPacket() []rtcp.Packet {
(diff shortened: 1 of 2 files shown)


Sean-Der:
2021-09-11-211955_1215x198_scrot

Sean-Der:
@mengelbart It works now! Amazing work :)

What does `tlcc` mean? Wasn't exactly sure so changed to `feedback` instead.

mengelbart:
Great! Thanks for fixing this! 

I think tlcc was meant short for `TransportLayerCC`, which is the type of the returned RTCP packet, but I agree `feedback` is much better!

mengelbart (approved):

PR #67, opened 2021-09-16
Add InterceptorFactory

Sean-Der:
Interceptors are being accidentally misused by users.
The issue is that an Interceptor can be re-used between
multiple PeerConnections. Interceptors were designed to only be
single PeerConnection aware, so state is being corrupted.

Instead we are now going to provide InterceptorFactories. The default
API of pion/webrtc will now be safe to use between PeerConnections.

Relates to pion/webrtc#1956

### DIFF
--- a/examples/nack/main.go
+++ b/examples/nack/main.go
@@ -35,7 +35,12 @@ func receiveRoutine() {
 	}
 
 	// Create NACK Generator
-	generator, err := nack.NewGeneratorInterceptor()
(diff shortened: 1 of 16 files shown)


Sean-Der:
Do people prefer that `Option` gets executed once per factory or interceptor?

I opted for once per Interceptor
* You can customize each one
* You could track interceptor creation

It could be wasteful resource wise, but I don't believe enough to matter.

davidzhao (approved):
this looks good!

Issue #25, opened 2020-12-30
Bandwidth Estimator


Sean-Der:
This is my rough plan for Bandwidth Estimation @aler9. I could totally be missing something, but just been my research so far.

WebRTC uses [TWCC](https://tools.ietf.org/html/draft-holmer-rmcat-transport-wide-cc-extensions-01) for feedback. This is used instead of receiver reports because it can also communicate queuing. We should have an Interceptor that emits these reports for incoming packets.

Chrome then uses [GCC]( https://tools.ietf.org/html/draft-ietf-rmcat-gcc-02) we should have an interceptor that consumes the TWCC messages and emits a value for the user.

I have also seen some people exploring [BBR](https://tools.ietf.org/html/draft-cardwell-iccrg-bbr-congestion-control-00) instead. It is easier to find information on this implementation. I don't know how it performs head-to-head, but it might be easier to start with this. Already seems to be BBR(like?) implementations in Go https://github.com/go-kratos/kratos/blob/master/pkg/ratelimit/bbr/bbr.go

Sean-Der:
https://gstconf.ubicast.tv/videos/google-transport-wide-congestion-control/ is a good talk on TWCC, it is available in GStreamer today which could be useful for our debugging/writing. 

We also already have an implementation in Ion. https://github.com/pion/ion-sfu/blob/3ae96731c273ab2d459e9c42793d928feec0a84f/pkg/buffer/twcc.go

Sean-Der:
I reached out to ibc and the issue is that GCC is not fully documented. The GCC draft I posted above misses a bunch of important details.

With that in mind I think BBR is the best path to go down! This is my high level list of items.

----
* [ ] TWCC Generator
* [ ] Extend VNet so we can simulate packet loss and packet delay
* [ ] BBR Controller. It should be decoupled from WebRTC (possibly ship as its own package?) could be useful for others
* [ ] BBR Interceptor. Parse incoming TWCC RTCP messages and interface with BBR Controller via its Public API.

jech:
Accurate SR/RR is the most urgent bit — without that, nothing works correctly.  SR are essential for correct lipsynch, and GCC will yield acceptable performance on non-bufferbloated networks with just correct RR.

Let me expand on that.  GCC consists of two congestion controllers:

  - a fairly traditional loss-based controller, that reacts to the loss rate communicated in RR;
  - a rather novel delay-based controller, that reacts to the second derivative of the packet arrival time.

[...]

PR #71, opened 2021-11-02
[feat/gcc] Implement Bandwidth Estimation

mengelbart:
#### Description

This pull request tracks the implementation progress for bandwidth estimation using Google Congestion Control. It will introduce interfaces that can hopefully be reused by other bandwidth estimators in the future.

#### Reference issue
Implements the GCC related tasks of #25

#### TODO:
- [x] Implement TWCC Feedback adapter
- [x] Integrate in bwe tests in pion/webrtc
- [x] Sending Engine (pacer)
- [ ] Implement Loss based bandwidth estimation
  - [x] Calculate loss and update target bitrate
  - [ ] `loss_based_bwe.go` unit tests
- [x] Implement Delay based bandwidth estimation
  - [x] Pre-filter acknowledgements and aggregate arrival groups
  - [x] Implement arrival-time filter to estimate inter group delay variation
  - [x] Implement over-use-detector using adaptive threshold
  - [x] Implement Rate Control
  - [x] Add measurement uncertainty updates to kalman filter
  - [x] Add outlier filter for non-zero-mean WGN of measurement uncertainty
  - [x] Add over use time threshold
  - [x] Tune parameters (adaptive threshold, ...)
- [ ] (optional/later) Implement RFC8888
- [ ] (optional/later) Implement RFC8888 Feedback adapter

### DIFF
--- a/pkg/cc/interceptor.go
+++ b/pkg/cc/interceptor.go
@@ -1,0 +1,147 @@
+// Package cc implements an interceptor for bandwidth estimation that can be
+// used with different BandwidthEstimators.
+package cc
+
+import (
+	"github.com/pion/interceptor"
+	"github.com/pion/interceptor/pkg/gcc"
+	"github.com/pion/logging"
+	"github.com/pion/rtcp"
+)
+
+const transportCCURI = "http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01"
+
+// Option can be used to set initial options on CC interceptors
+type Option func(*Interceptor) error
+
+// BandwidthEstimatorFactory creates new BandwidthEstimators
+type BandwidthEstimatorFactory func() (BandwidthEstimator, error)
+
+// BandwidthEstimator is the interface that will be returned by a
+// NewPeerConnectionCallback and can be used to query current bandwidth
+// metrics and add feedback manually.
+type BandwidthEstimator interface {
+	AddStream(*interceptor.StreamInfo, interceptor.RTPWriter) interceptor.RTPWriter
+	WriteRTCP([]rtcp.Packet, interceptor.Attributes) error
+	GetTargetBitrate() int
+	OnTargetBitrateChange(f func(bitrate int))
+	GetStats() map[string]interface{}
+	Close() error
+}
+
+// NewPeerConnectionCallback returns the BandwidthEstimator for the
+// PeerConnection with id
+type NewPeerConnectionCallback func(id string, estimator BandwidthEstimator)
+
+// InterceptorFactory is a factory for CC interceptors
+type InterceptorFactory struct {
+	opts              []Option
+	bweFactory        func() (BandwidthEstimator, error)
+	addPeerConnection NewPeerConnectionCallback
+}
+
+// NewInterceptor returns a new CC interceptor factory
+func NewInterceptor(factory BandwidthEstimatorFactory, opts ...Option) (*InterceptorFactory, error) {
+	if factory == nil {
+		factory = func() (BandwidthEstimator, error) {
+			return gcc.NewSendSideBWE()
+		}
+	}
+	return &InterceptorFactory{
+		opts:              opts,
+		bweFactory:        factory,
+		addPeerConnection: nil,
+	}, nil
+}
+
+// OnNewPeerConnection sets a callback that is called when a new CC interceptor
+// is created.
+func (f *InterceptorFactory) OnNewPeerConnection(cb NewPeerConnectionCallback) {
+	f.addPeerConnection = cb
+}
+
+// NewInterceptor returns a new CC interceptor
+func (f *InterceptorFactory) NewInterceptor(id string) (interceptor.Interceptor, error) {
+	bwe, err := f.bweFactory()
+	if err != nil {
+		return nil, err
+	}
+	i := &Interceptor{
+		NoOp:      interceptor.NoOp{},
+		log:       logging.NewDefaultLoggerFactory().NewLogger("cc_interceptor"),
+		estimator: bwe,
+		feedback:  make(chan []rtcp.Packet),
+		close:     make(chan struct{}),
+	}
+
+	for _, opt := range f.opts {
+		if err := opt(i); err != nil {
+			return nil, err
+		}
+	}
+
+	if f.addPeerConnection != nil {
+		f.addPeerConnection(id, i.estimator)
+	}
+	return i, nil
+}
+
+// Interceptor implements Google Congestion Control
+type Interceptor struct {
+	interceptor.NoOp
+	log       logging.LeveledLogger
+	estimator BandwidthEstimator
+	feedback  chan []rtcp.Packet
+	close     chan struct{}
+}
+
+// BindRTCPReader lets you modify any incoming RTCP packets. It is called once
+// per sender/receiver, however this might change in the future. The returned
+// method will be called once per packet batch.
+func (c *Interceptor) BindRTCPReader(reader interceptor.RTCPReader) interceptor.RTCPReader {
+	return interceptor.RTCPReaderFunc(func(b []byte, a interceptor.Attributes) (int, interceptor.Attributes, error) {
+		i, attr, err := reader.Read(b, a)
+		if err != nil {
+			return 0, nil, err
+		}
+		buf := make([]byte, i)
+
+		copy(buf, b[:i])
+
+		if attr == nil {
+			attr = make(interceptor.Attributes)
+		}
+
+		pkts, err := attr.GetRTCPPackets(buf[:i])
+		if err != nil {
+			return 0, nil, err
+		}
+		if err = c.estimator.WriteRTCP(pkts, attr); err != nil {
+			return 0, nil, err
+		}
+
+		return i, attr, nil
+	})
+}
+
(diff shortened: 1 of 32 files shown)


pkg/cc/interceptor.go
```diff
@@ -0,0 +1,152 @@
...
+
+	"github.com/pion/interceptor"
+	"github.com/pion/interceptor/pkg/gcc"
+	"github.com/pion/logging"
+	"github.com/pion/rtcp"
+)
+
+const transportCCURI = "http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01"
+
+// ErrUnknownSession indicates that a session ID was not assigned
```
Sean-Der, line 16: This is unused, can we delete?

pkg/gcc/adaptive_threshold.go
```diff
@@ -0,0 +1,60 @@
...
+	overuseCoefficientU float64
+	overuseCoefficientD float64
+	min                 time.Duration
+	max                 time.Duration
+}
+
+func newAdaptiveThreshold(opts ...adaptiveThresholdOption) *adaptiveThreshold {
+	at := &adaptiveThreshold{
+		thresh:              time.Duration(125 * float64(time.Microsecond)),
+		overuseCoefficientU: 0.01,
```
Sean-Der, line 26: Where do these values come from? Would it be possible to link to a RFC/PDF?

Maybe we could write an informational one if not?

pkg/gcc/adaptive_threshold.go
```diff
@@ -0,0 +1,60 @@
...
+
+type adaptiveThreshold struct {
+	thresh              time.Duration
+	overuseCoefficientU float64
+	overuseCoefficientD float64
+	min                 time.Duration
+	max                 time.Duration
+}
+
+func newAdaptiveThreshold(opts ...adaptiveThresholdOption) *adaptiveThreshold {
```
Sean-Der, line 23: Could you write a short `godoc` on what a AdaptiveThreshold does? Nothing crazy, just to re-inforce what it does to someone reading the code?

pkg/gcc/rate_controller.go
```diff
@@ -0,0 +1,177 @@
...
+	initialTargetBitrate int
+	minBitrate           int
+	maxBitrate           int
+
+	lastUpdate         time.Time
+	target             int
+	lastState          state
+	latestRTT          time.Duration
+	latestReceivedRate int
+	decreaseEMA        float64
```
Sean-Der, line 22: What does EMA stand for? Maybe worth expanding

pkg/gcc/state.go
```diff
@@ -0,0 +1,59 @@
+package gcc
+
+import "fmt"
+
+type state int
+
+const (
+	increase state = iota
```
Sean-Der, line 8: To prevent name collisions I usually prefix enum.

```
stateIncrease
stateDecrease
stateHold
```

pkg/gcc/static_threshold.go
```diff
@@ -0,0 +1,17 @@
+package gcc
```
Sean-Der, line 1: Should this be moved into `overuse_detector_test.go` I don't see it used anywhere else? Do we plan on using it more/is it a common idiom?

pkg/gcc/send_side_bwe.go
```diff
@@ -0,0 +1,201 @@
...
+// Stats contains internal statistics of the bandwidth estimator
+type Stats struct {
+	LossStats
+	DelayStats
+}
+
+// SendSideBWE implements a combination of loss and delay based GCC
+type SendSideBWE struct {
+	pacer Pacer
+	lc    *lossBasedBandwidthEstimator
```
Sean-Der, line 31: IMO code is easier to read with longer names on these. `lossController` `delayController` and `feedbackAdapter` had to scroll back a few times to find myself again.

pkg/gcc/loss_based_bwe.go
```diff
@@ -0,0 +1,93 @@
...
+	defer e.lock.Unlock()
+
+	if increaseLoss < 0.02 && time.Since(e.lastIncrease) > 200*time.Millisecond {
+		e.log.Infof("loss controller increasing; averageLoss: %v, decreaseLoss: %v, increaseLoss: %v", e.averageLoss, decreaseLoss, increaseLoss)
+		e.lastIncrease = time.Now()
+		e.bitrate = clampInt(int(1.05*float64(e.bitrate)), e.minBitrate, e.maxBitrate)
+	} else if decreaseLoss > 0.1 && time.Since(e.lastDecrease) > 200*time.Millisecond {
+		e.log.Infof("loss controller decreasing; averageLoss: %v, decreaseLoss: %v, increaseLoss: %v", e.averageLoss, decreaseLoss, increaseLoss)
+		e.lastDecrease = time.Now()
+		e.bitrate = clampInt(int(float64(e.bitrate)*(1-0.5*decreaseLoss)), e.minBitrate, e.maxBitrate)
```
Sean-Der, line 87: Why `1-0.5` and not `0.95` ?

pkg/gcc/loss_based_bwe.go
```diff
@@ -0,0 +1,93 @@
...
+	e.averageLoss = e.average(time.Since(e.lastLossUpdate), e.averageLoss, lossRatio)
+	e.lastLossUpdate = time.Now()
+
+	increaseLoss := math.Max(e.averageLoss, lossRatio)
+	decreaseLoss := math.Min(e.averageLoss, lossRatio)
+
+	e.lock.Lock()
+	defer e.lock.Unlock()
+
+	if increaseLoss < 0.02 && time.Since(e.lastIncrease) > 200*time.Millisecond {
```
Sean-Der, line 80: I would make the thresholds + time constants `0.02` and `200 * time.Millisecond`  Also might be worth linking the PDF I remember seeing the constants in there?

pkg/gcc/acknowledgment.go
```diff
@@ -0,0 +1,26 @@
+package gcc
```
Sean-Der, line 1: Do you think we could move this to `internal/cc`

This seems like something we could share between congestion controllers. Would love to see this directory be dedicated to gcc specific stuff!

mengelbart: Yes, I think we could move it out of this package, but I'm not sure what the right place would be, `internal/cc` would mean that `pkg/gcc` can basically only be used from this module, right? I think it would be nice to allow using `pkg/gcc` without interceptors?

pkg/gcc/rate_controller.go
mengelbart: Exponential moving average, I moved it into a separate struct, this should make it clearer, also how it is calculated.

pkg/cc/interceptor.go
mengelbart: Yes, done, thanks!

pkg/gcc/adaptive_threshold.go
mengelbart: I made some more changes to the rate controller and adaptive threshold and added some explanations.

pkg/gcc/acknowledgment.go
mengelbart: Yes, I think that would work, but currently I use a default `BandwidthEstimator` in the interceptor, so that would create an import cycle. But maybe there's a better way to set the default estimator without including the package just for that?

mengelbart: *and the default uses `gcc.NewSendSideBWE()`

Sean-Der: We could maybe do `twcc` for now and fix in the future?

Sean-Der (approved):

pkg/gcc/kalman.go
```diff
@@ -0,0 +1,91 @@
...
+	return k
+}
+
+func (k *kalman) updateEstimate(measurement time.Duration) time.Duration {
+	z := measurement - k.estimate
+
+	zms := float64(z.Microseconds()) / 1000.0
+
+	if !k.disableMeasurementUncertaintyUpdates {
+		alpha := math.Pow((1 - chi), 30.0/(1000.0*5*float64(time.Millisecond)))
```
alexpokotilo, line 75: @mengelbart 
I was not able to get why we use "5*float64(time.Millisecond)"
for f_max in 
alpha = (1-chi)^(30/(1000 * f_max)) calculation.
This is important part and has impact on uncertainty calculation. Please help to clue

pkg/gcc/adaptive_threshold.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return at
+}
+
+func (a *adaptiveThreshold) compare(estimate, dt time.Duration) (usage, time.Duration, time.Duration) {
+	a.numDeltas++
+	if a.numDeltas < 2 {
+		return usageNormal, estimate, a.max
+	}
+	t := time.Duration(minInt(a.numDeltas, maxDeltas)) * estimate
```
alexpokotilo, line 65: @mengelbart 
why do we multiply estimate(m_hat(i)) from Kalman filter by up to 60 ?
I was not able to find this in standard and/or https://c3lab.poliba.it/images/6/65/Gcc-analysis.pdf

PR #74, opened 2021-11-18
Update interceptor API to use Reader/Writer-style interfaces

kevmo314:
This change updates the interceptor API to use a reader/writer interface. The best example of use is in `examples/nack/main.go` which is substantially cleaner with this API.

The primary API is under `pkg/rtpio` and closely mirrors that of `io.Reader` and `io.Writer`. Essentially, there are two types of interceptors:

```
Sender interceptors:

                +----------------+
                |                | -> RTP Writer  ->
 RTP Writer ->  |   interceptor  | -> RTCP Writer ->
                |                |
                +----------------+
                         ^
                         |
        --------->  RTCP Reader  <---------------

        
Receiver interceptors:

                +----------------+
 RTCP Writer <- |                | 
                |   interceptor  | -> RTP Reader ->
 RTP Reader ->  |                |
                +----------------+
                         ^
                         |
        --------->  RTCP Reader  <---------------
```

With the `rtpio` primitives it's sufficiently flexible to compose all of them.

It's not lost on me that this is a massive change, hence the migration to a v2 module. Unfortunately I'm not sure of a good way to piecemeal the change together. Fortunately, all the tests are still in place and there should be no behavior changes and hopefully the 800 lines of code reduction is enticing :)


kevmo314:
@Sean-Der any thoughts on the API here?

Sean-Der:
Really great work @kevmo314! This reduces the complexity a lot, it is so exciting to have someone deep in the interceptors. Here are some asks that came up with the v0.


### InterceptoryFactory/Opt-in Global State

Some users want to use one thread to serve multiple Interceptors. The `InterceptorFactory` pattern lets users have a Factory that can service multiple interceptors. We wanted to make it possible to spawn a thread in the factory and have it service multiple interceptors.

### Copying 

Copying for each RTP/RTCP Read is going to make things more expensive. I think interceptors make sense not to copy because we have lots of small operations. If a individual interceptor needs to keep a copy of a packet they should do that themselves.

### Backlog

backlog could be a problem. If an attacker knows that you are running Pion then can send RTCP packets with 17 packets and grow that buffer forever. Maybe just need to set an upper limit?

PR #76, opened 2021-11-22
Reduce NACK buffer size to 512

boks1971:
Bug
https://github.com/pion/webrtc/issues/2028


pkg/nack/receive_log_test.go
```diff
@@ -8,8 +8,6 @@ import (
 
 func TestReceivedBuffer(t *testing.T) {
 	for _, start := range []uint16{0, 1, 127, 128, 129, 511, 512, 513, 32767, 32768, 32769, 65407, 65408, 65409, 65534, 65535} {
-		start := start
```
boks1971, line 11: looked redundant. So, removed it.

Sean-Der (approved):

PR #77, opened 2021-11-22
Fix assignment to entry in nil map

mengelbart:
We started using attributes to store headers instead of repeatedly
parsing them in every interceptor. This only works if the attributes are
initialized before they are used.

### DIFF
--- a/pkg/nack/receive_log_test.go
+++ b/pkg/nack/receive_log_test.go
@@ -8,6 +8,8 @@ import (
 
 func TestReceivedBuffer(t *testing.T) {
 	for _, start := range []uint16{0, 1, 127, 128, 129, 511, 512, 513, 32767, 32768, 32769, 65407, 65408, 65409, 65534, 65535} {
+		start := start
+
 		t.Run(fmt.Sprintf("StartFrom%d", start), func(t *testing.T) {
 			rl, err := newReceiveLog(128)
 			if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/nack/receive_log_test.go
```diff
@@ -8,6 +8,8 @@ import (
 
 func TestReceivedBuffer(t *testing.T) {
 	for _, start := range []uint16{0, 1, 127, 128, 129, 511, 512, 513, 32767, 32768, 32769, 65407, 65408, 65409, 65534, 65535} {
+		start := start
```
boks1971, line 11: Looks good @mengelbart . I had removed this line earlier in the day in another PR as it looked redundant. Is this needed?

boks1971 (approved):

pkg/nack/receive_log_test.go
mengelbart: I think it depends on the test case: `start` will be overwritten in each iteration, so if the function called in the loop uses `start` after the iteration is finished, it will operate on the value of the next iteration. Using `start := start` copies the value to a new variable. Not sure if this is a real issue here, but better be safe and `golangci-lint` will complain about it.

boks1971: Got it. I removed it and the tests passed. So thought it was okay to remove.

Sean-Der (approved):

PR #86, opened 2021-12-19
Avoid a copy for NoOpPacer and use buffer pool

boks1971:
#### Description
- Call c.pacer.Write directly. Introduces some delay by putting packet
  into channel. This also avoids a copy of packet when using NoOpPacer.
- Use a buffer pool in leaky_bucket_pacer.go, potentially cheaper to
  re-use buffers.
- Do not fail if packet does not have CC extension. This allows use of
  TWCC with just one kind of media type. It is mostly used only with
  video.
- Introduce a callback in BandwidthEstimator interface to get notified
  on target bitrate changes.

### DIFF
--- a/pkg/gcc/interceptor.go
+++ b/pkg/gcc/interceptor.go
@@ -44,9 +44,12 @@ func SetPacer(pacer Pacer) Option {
 }
 
 type BandwidthEstimator interface {
+	AddFeedback(cc *rtcp.TransportLayerCC)
+
 	GetTargetBitrate() int
-	GetStats() map[string]interface{}
 	OnTargetBitrateChange(f func(bitrate int))
+
+	GetStats() map[string]interface{}
 }
 
 type NewPeerConnectionCallback func(id string, estimator BandwidthEstimator)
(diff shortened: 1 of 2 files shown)


pkg/gcc/interceptor.go
```diff
@@ -195,7 +194,7 @@ func (c *Interceptor) BindLocalStream(info *interceptor.StreamInfo, writer inter
 
 	c.pacer.AddStream(info.SSRC, interceptor.RTPWriterFunc(func(header *rtp.Header, payload []byte, attributes interceptor.Attributes) (int, error) {
 		// Call adapter.onSent
-		if err := c.OnSent(time.Now(), header, len(payload), attributes); err != nil {
+		if err := c.OnSent(time.Now(), header, len(payload), attributes); err != nil && err != errMissingTWCCExtension {
```
mengelbart, line 204: I have a question about this: Does it make sense to send the stream through this interceptor at all? Could we just do:
```
	var hdrExtID uint8
	for _, e := range info.RTPHeaderExtensions {
		if e.URI == transportCCURI {
			hdrExtID = uint8(e.ID)
			break
		}
	}
	if hdrExtID == 0 {
		return writer
	}

	c.pacer.AddStream(info.SSRC, interceptor.RTPWriterFunc(func(header *rtp.Header, payload []byte, attributes interceptor.Attributes) (int, error) {
```
instead, to bypass the whole interceptor? (header extension IDs can't be 0).

PR #88, opened 2021-12-20
Bypass GCC interceptor for stream not using TWCC

boks1971:
- Simplify a couple of other bits of code.

### DIFF
--- a/pkg/gcc/leaky_bucket_pacer.go
+++ b/pkg/gcc/leaky_bucket_pacer.go
@@ -123,7 +123,7 @@ func (p *LeakyBucketPacer) Run() {
 					p.log.Errorf("failed to write packet: %v", err)
 				}
 				p.pool.Put(next.payload)
-				budget -= int64(n)
+				budget -= int64(n * 8)
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


pkg/gcc/leaky_bucket_pacer.go
```diff
@@ -123,7 +123,7 @@ func (p *LeakyBucketPacer) Run() {
 					p.log.Errorf("failed to write packet: %v", err)
 				}
 				p.pool.Put(next.payload)
-				budget -= int64(n)
+				budget -= int64(n * 8)
```
boks1971, line 126: @mengelbart Not sure if this correct. I think `Write` returns bytes and `targetBitrate` and `budget` are in bits. So, multiplying what is returned by `Write` by 8 for budget adjustment.

boks1971: On a related note, would be good to maintain state about last send loop time vs current time and log some oddities of timer wake up. At scale, guessing this could add up. Imagine a server running a 1000 of these. At 5ms, that is 200K wake ups and processing per second. So, might be worth logging about things lagging.

mengelbart: Looks correct to me, thanks!

mengelbart: Yep, I agree, we should do some tests under high load.

PR #92, opened 2022-01-06
Fix NACK interceptor sendBuffer overridden

### LINKED ISSUE #84: NACK responder always responding with most recent packet
In the NACK responder, it tries to cache sent packets in order to respond to NACKs. However, it's storing a pointer to the rtp.Packet struct, which may be re-used by the sender process without the knowledge of NackResponder.

Adam from the community has observed symptoms where retransmitted packets contained payload that did not match the sequence number that was requested. (but a later sequence number that overrode the earlier one).

We need to be storing a copy of the packet to ensure that nack responder has full ownership over its data.

davidzhao:
Keeps a copy of packet in responder_interceptor, fixes #84

While trying to fix this, I was also thinking about how we might reduce copy/allocations throughout pion. Have implemented a version of this that works within intercepters (unexported for now). any feedback is welcome!

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -9,6 +9,7 @@ aler9 <46489434+aler9@users.noreply.github.com>
 Antoine Baché <antoine@tenten.app>
 Atsushi Watanabe <atsushi.w@ieee.org>
 boks1971 <raja.gobi@tutanota.com>
+David Zhao <david@davidzhao.com>
 Jonathan Müller <jonathan@fotokite.com>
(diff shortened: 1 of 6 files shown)


davidzhao:
Actually, just realized this doesn't cover header extensions being overridden. Will need to marshal/unmarshal for now. Would be great to use @kevmo314's zero copy headers as well.

kevmo314 (approved):
Looks good to me, yeah would love to get the zero copy stuff in too :)

PR #98, opened 2022-01-21
Add RFC8888 packet stream implementation

kevmo314:
Taken from https://github.com/muxable/rtptools. These structs implement RFC8888 and allow congestion control to operate purely on the sender.

We'll need to write a separate interceptor to emit these, but in this change there is no actual link to an interceptor only the packet construction.


aler9:
Hello, i don't think it would be wise to add external dependencies - in particular, the new dependency doesn't even have a license, and this would cause a lot of problems for both documentation (go.pkg.dev refuses to serve documentation of packages without license) and users (there are a lot of users that needs to do dependency check to ensure that their libraries are compatible with their intended aim).

In my opinion you'd better contribute your entire code to this library.

mengelbart:
This was integrated in #112, thanks @kevmo314 for preparing it!

PR #109, opened 2022-03-12
Move sender report generation into sender_stream.go

aler9:
#### Description

In this way, receiver and sender report generators are aligned


Sean-Der:
LGTM! Thanks for the refactor @aler9 :)

PR #124, opened 2022-04-19
Fix lock after closing interceptor

braams:
When we close s.closed channel we stop reading from s.packetChan channel in loop function. But we still continue writing to s.packetChan channel in RTPReaderFunc. Sometimes it causes lock on reading from track. This PR aims to fix it.


Sean-Der:
Nice catch! Thank you so much @braams

Sean-Der:
Would you mind adding a subject to your commit message? I would just copy what you said on Github.

You can test if it worked with `bash .github/lint-commit-message.sh`

PR #132, opened 2022-05-20
fix(ss-bwe): `WriteRTCP` raise panic after `Close`

### LINKED ISSUE #131: [SendSideBWE ] `delayController` raises panic when packet arrives after interceptor is closed
### Your environment.
  - Version: `github.com/pion/interceptor v0.1.11-0.20220401022307-09051cdde82b`
  - Other Information - 
 
Panic stack trace:
```
runtime.gopanic
        /usr/local/go/src/runtime/panic.go:965
runtime.chansend
        /usr/local/go/src/runtime/chan.go:281
runtime.chansend1
        /usr/local/go/src/runtime/chan.go:143
github.com/pion/interceptor/pkg/gcc.(*delayController).updateDelayEstimate
        /go/pkg/mod/github.com/pion/interceptor@v0.1.11-0.20220401022307-09051cdde82b/pkg/gcc/delay_based_bwe.go:96
github.com/pion/interceptor/pkg/gcc.(*SendSideBWE).WriteRTCP
        /go/pkg/mod/github.com/pion/interceptor@v0.1.11-0.20220401022307-09051cdde82b/pkg/gcc/send_side_bwe.go:141
github.com/100mslive/ion-sfu/pkg/sfu.(*TrackAllocator).OnTransportCCFeedback
        /go/pkg/mod/github.com/100mslive/ion-sfu@v1.9.9-0.20220510151633-ae293ba3929d/pkg/sfu/track_allocator.go:459
github.com/100mslive/ion-sfu/pkg/sfu.(*DownTrack).handleRTCP
        /go/pkg/mod/github.com/100mslive/ion-sfu@v1.9.9-0.20220510151633-ae293ba3929d/pkg/sfu/downtrack.go:985
github.com/100mslive/ion-sfu/pkg/sfu.(*DownTrack).Bind.func1
        /go/pkg/mod/github.com/100mslive/ion-sfu@v1.9.9-0.20220510151633-ae293ba3929d/pkg/sfu/downtrack.go:273
github.com/100mslive/ion-sfu/pkg/buffer.(*RTCPReader).Write
        /go/pkg/mod/github.com/100mslive/ion-sfu@v1.9.9-0.20220510151633-ae293ba3929d/pkg/buffer/rtcpreader.go:25
github.com/pion/srtp/v2.(*ReadStreamSRTCP).write
        /go/pkg/mod/github.com/pion/srtp/v2@v2.0.5/stream_srtcp.go:30
github.com/pion/srtp/v2.(*SessionSRTCP).decrypt
        /go/pkg/mod/github.com/pion/srtp/v2@v2.0.5/session_srtcp.go:173
github.com/pion/srtp/v2.(*session).start.func1
        /go/pkg/mod/github.com/pion/srtp/v2@v2.0.5/session.go:141

```
 
### What did you expect?
No panics

### What happened?
There's a race between RTCP packets written to `SendSideBWE` and interceptor being closed due to which any packets written after the interceptor is closed raised the panic `send on closed channel`. This panic happens very rarely (1 out of 200 `Peer`) which makes it pretty hard to reproduce in local env. It happens at the same time when ice state goes into `closed` state and respective peer-connection is closed 

Would ignoring any new packets after the interceptor is closed be a good fix? @Sean-Der

adityaa30:
#### Description
Return an error when `WriteRTCP` is called after the `SendSideBWE` is closed

#### Reference issue
Fixes #131

### DIFF
--- a/pkg/gcc/send_side_bwe.go
+++ b/pkg/gcc/send_side_bwe.go
@@ -50,7 +50,8 @@ type SendSideBWE struct {
 	minBitrate    int
 	maxBitrate    int
 
-	close chan struct{}
+	close     chan struct{}
+	closeLock sync.RWMutex
 }
 
 // Option configures a bandwidth estimator
@@ -135,7 +136,10 @@ func (e *SendSideBWE) AddStream(info *interceptor.StreamInfo, writer interceptor
 
 // WriteRTCP adds some RTCP feedback to the bandwidth estimator
 func (e *SendSideBWE) WriteRTCP(pkts []rtcp.Packet, attributes interceptor.Attributes) error {
-	if e.IsClosed() {
+	e.closeLock.RLock()
+	defer e.closeLock.RUnlock()
+
+	if e.isClosed() {
 		return ErrSendSideBWEClosed
 	}
 
@@ -184,8 +188,8 @@ func (e *SendSideBWE) OnTargetBitrateChange(f func(bitrate int)) {
 	e.onTargetBitrateChange = f
 }
 
-// IsClosed returns true if SendSideBWE is closed
-func (e *SendSideBWE) IsClosed() bool {
+// isClosed returns true if SendSideBWE is closed
+func (e *SendSideBWE) isClosed() bool {
 	select {
 	case <-e.close:
 		return true
@@ -196,6 +200,9 @@ func (e *SendSideBWE) IsClosed() bool {
 
 // Close stops and closes the bandwidth estimator
 func (e *SendSideBWE) Close() error {
+	e.closeLock.Lock()
+	defer e.closeLock.Unlock()
+
 	if err := e.delayController.Close(); err != nil {
 		return err
 	}
--- a/pkg/gcc/send_side_bwe.go
+++ b/pkg/gcc/send_side_bwe.go
@@ -1,6 +1,7 @@
 package gcc
 
 import (
+	"errors"
 	"sync"
 	"time"
 
@@ -16,6 +17,9 @@ const (
 	minBitrate     = 5_000
 	maxBitrate     = 50_000_000
 )
+
+// ErrSendSideBWEClosed is raised when SendSideBWE.WriteRTCP is called after SendSideBWE.Close
+var ErrSendSideBWEClosed = errors.New("SendSideBwe closed")
 
 // Pacer is the interface implemented by packet pacers
 type Pacer interface {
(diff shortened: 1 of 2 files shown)


pkg/gcc/send_side_bwe.go
```diff
@@ -131,6 +136,14 @@ func (e *SendSideBWE) AddStream(info *interceptor.StreamInfo, writer interceptor
 
 // WriteRTCP adds some RTCP feedback to the bandwidth estimator
 func (e *SendSideBWE) WriteRTCP(pkts []rtcp.Packet, attributes interceptor.Attributes) error {
+	e.lock.RLock()
```
mengelbart, line 139: I would suggest to use something like this here:
```
select {
    case <-e.close:
        return ErrSendSideBWEClosed
    default:
}
```
to save the boolean here.

adityaa30: Thanks! I've updated the PR

pkg/gcc/send_side_bwe.go
```diff
@@ -131,6 +135,10 @@ func (e *SendSideBWE) AddStream(info *interceptor.StreamInfo, writer interceptor
 
 // WriteRTCP adds some RTCP feedback to the bandwidth estimator
 func (e *SendSideBWE) WriteRTCP(pkts []rtcp.Packet, attributes interceptor.Attributes) error {
+	if e.IsClosed() {
```
mengelbart, line 138: I think we still need the lock to make sure the interceptor doesn't get closed while looping over `pkts`. Maybe we should use a different lock that only locks access to `e.close` and does not block all the other places that currently use `e.lock`.

pkg/gcc/send_side_bwe.go
```diff
@@ -176,6 +184,16 @@ func (e *SendSideBWE) OnTargetBitrateChange(f func(bitrate int)) {
 	e.onTargetBitrateChange = f
 }
 
+// IsClosed returns true if SendSideBWE is closed
+func (e *SendSideBWE) IsClosed() bool {
```
mengelbart, line 188: Do we need to export `IsClosed` or is it enough to have it in this package?

mengelbart (changes requested):

pkg/gcc/send_side_bwe.go
adityaa30: Done

pkg/gcc/send_side_bwe.go
adityaa30: Thinking about it again, it should be enough to keep it inside the package as other `isClosed` implementations are also not exposed.

mengelbart (approved):
LGTM! Thanks for fixing this!

PR #137, opened 2022-05-24
Support publishing average loss rate

kevmo314:
This change exposes an API for reading the loss rate.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -14,5 +14,6 @@ boks1971 <raja.gobi@tutanota.com>
 David Zhao <david@davidzhao.com>
 Jonathan Müller <jonathan@fotokite.com>
 Kevin Caffrey <kcaffrey@gmail.com>
+Kevin Wang <kevmo314@gmail.com>
 Mathis Engelbart <mathis.engelbart@gmail.com>
 Sean DuBois <sean@siobud.com>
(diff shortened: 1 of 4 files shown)


mengelbart:
Hi @kevmo314
sorry I didn't reply to this earlier. I don't think we should add the loss rate to the `BandwidthEstimator` interface, because it is not clear if all implementations will use or even know about packet loss. Also, the implementation of the GCC estimator currently does not really reflect how the RTP packet loss rate should be calculated.

I just started work on a new interceptor to keep track of all kinds of RTP and RTCP-related statistics in #143 which includes loss rate in a cleaner implementation. Would that interceptor be a good alternative for your use case?

kevmo314:
> Hi @kevmo314 sorry I didn't reply to this earlier. I don't think we should add the loss rate to the `BandwidthEstimator` interface, because it is not clear if all implementations will use or even know about packet loss. Also, the implementation of the GCC estimator currently does not really reflect how the RTP packet loss rate should be calculated.
> 
> I just started work on a new interceptor to keep track of all kinds of RTP and RTCP-related statistics in #143 which includes loss rate in a cleaner implementation. Would that interceptor be a good alternative for your use case?

Seems reasonable, but aren't SR's sent at much lower frequency? The loss rate in our use case gets piped to the Opus/FEC encoder so something that's sent at 1 Hz might be too slow to respond to a sudden increase in loss.

mengelbart:
Mh, interesting, no idea if that will be helpful for you then. Anyway, I think the GCC interceptor allows you to get `averageLoss` via the `GetStats`, too, but I think that may change at some point or at least the way the value is calculated may change in the future.
If you just want full control of the congestion controller, you can also pass your own `BandwidthEstimator` implementation, e.g. by forking the one in pkg/gcc and adding/removing whatever you need :)

Issue #138, opened 2022-06-02
History in cc's `FeedbackAdapter` is never cleaned up


rlpeck:
### Your environment.
  - Version: current
  - Browser: N/A
  - Other Information - N/A

### What did you do?

I noticed that the `history` map in [`cc/feedback_adapter.go`](https://github.com/pion/interceptor/blob/master/internal/cc/feedback_adapter.go#L27) never has elements removed. That means that map could grow to have a size of `1<<16 - 1` for each track where this is used.

### What did you expect?

I would have expected to see some form of cleanup of the `history` map but I'm not sure if time based makes sense.

PR #140, opened 2022-06-07
Reduce the Upper bound on Feedback History size

rlpeck:
#### Description
Puts a smaller upper bound  of 250 on  number of `Acknowledgment`s that are stored in the `FeedbackAdapter`. This upper bound is completely arbitrary. Previously the number of `Acknowledgement`s was bound by the the max value of `uint16 (65535)`

#### Reference issue
Fixes https://github.com/pion/interceptor/issues/138

### DIFF
--- a/internal/cc/feedback_adapter.go
+++ b/internal/cc/feedback_adapter.go
@@ -1,6 +1,7 @@
 package cc
 
 import (
+	"container/list"
 	"errors"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


Sean-Der:
This is really great, thank you @rlpeck! Sorry I haven't made more movement on this. I would love to have you involved.

I am going to add you to the Pion organization now. Since you are actively using this I would love to talk about how we can make this better for everyone. I think we could make a bunch of great progress.

Sean-Der:
@rlpeck w/e you have the time reach out on video-dev or pion.ly/slack

mengelbart:
Sorry I didn't have time to check this out earlier. This looks good to me and we should definitely merge it, thank you @rlpeck for opening this PR!
Just one suggestion about the threshold: I quickly checked in chromium and they seem to be deleting packets from the history 60 seconds after they were created, so I suggest we use the same logic. But we can also merge this first and change it to a time-based threshold later or even make it configurable in case we want both.

Issue #93, opened 2022-01-13
Implement RFC 8888 feedback adapter for GCC


mengelbart:
The GCC implementation currently only works with TWCC feedback reports. It would be nice if it could also work with the feedback format described in [RFC 8888](https://www.rfc-editor.org/rfc/rfc8888.html). I think this should be straightforward to implement as it only needs to be added to `feedback_adapter.go`.

mengelbart:
Done in #145

PR #145, opened 2022-07-16
Add RFC8888 feedback to GCC

mengelbart:
Enables GCC to work with RFC8888 instead of TWCC.

Only merge after pion/rtcp is released with RFC 8888 support. Remove `replace` from `go.mod` before merging.


mengelbart:
I just realized this also depends on some functionality added by #143

mengelbart:
Closes #93

PR #148, opened 2022-07-16
Feat/gcc improved rtt approximation

mengelbart:
Uses a more accurate approximation of RTT by subtracting the time between packet arrival and feedback departure. Also removes RTT fields from each ACK because it was not used except for debugging.


mengelbart:
Draft because it depends on a number of previous PRs...

PR #151, opened 2022-07-29
feat: Add playout delay header interceptor

kevmo314:
This interceptor adds the playout delay header extension following
https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/playout-delay


pkg/playoutdelay/playout_delay.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// Extension is a extension payload format in
+// https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/playout-delay
+// 0                   1                   2                   3
+// 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |  ID   | len=2 |       MIN delay       |       MAX delay       |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+type Extension struct {
```
mengelbart, line 27: I think this should be part of pion/rtp

mengelbart (commented):
I'm not very familiar with the extension but have two questions: The [document](https://webrtc.googlesource.com/src/+/refs/heads/main/docs/native-code/rtp-hdrext/playout-delay) mentions that a sender may stop sending the extension if it receives a RR indicating that at it has been received at least once. Do we need that to save bytes?
And second question: Is it enough to send fixed limits, or do we need to be able to update the limits of the interceptor?

kevmo314:
> Do we need that to save bytes?

Yes, although since it's only a few bytes I figured I'd do it in a followup PR 😅 

> And second question: Is it enough to send fixed limits, or do we need to be able to update the limits of the interceptor?

The limits of the interceptor should be able to change but I'm actually not sure of a good way to do that. Since the developer only instantiates a factory, wouldn't they need access to the underlying interceptor to update the parameters on the fly? Do you have suggestions for such an API?

mengelbart:
> Yes, although since it's only a few bytes I figured I'd do it in a followup PR sweat_smile

Alright, follow up PR sounds fine to me :)

> The limits of the interceptor should be able to change but I'm actually not sure of a good way to do that. Since the developer only instantiates a factory, wouldn't they need access to the underlying interceptor to update the parameters on the fly? Do you have suggestions for such an API?

In the bandwidth estimation and similar interceptors, you can register a callback that is called whenever the factory creates a new interceptor. The callback receives the instance of the interceptor for the PeerConnection. I am not sure if this is the best design we can have but I don't know of a better way.

pkg/playoutdelay/playout_delay.go
kevmo314: Created https://github.com/pion/rtp/pull/196 to add it to pion/rtp.

PR #153, opened 2022-08-10
Update pion/rtp to v2

kevmo314:
Cleanup work to update to pion/rtp/v2.


mengelbart:
Do we need this (yet)? RTP `v2.0.0` doesn't seem to actually update anything and instead, just points to some old commit before `v.1.7.*`?

PR #163, opened 2022-11-29
Fix stats recorder actions that block when not yet started or stopped

asticode:
When closing a `PeerConnection` that is using the stats recorder as an interceptor, the process may block. I've dug a little bit and found out the calls to `GetStats()`, `QueueIncomingRTP()`, `QueueIncomingRTCP()`, `QueueOutgoingRTP()` and `QueueOutgoingRTCP()` block when the recorder is not yet started or stopped. I've added both `TestGetStatsNotBlocking` and `TestQueueNotBlocking` tests to test this.

I've proposed a fix which consists of replacing channels with mutexes but an alternative would be to wrap channel calls around a mutex.

Let me know what you think.
Cheers

### DIFF
--- a/pkg/stats/stats_recorder.go
+++ b/pkg/stats/stats_recorder.go
@@ -1,6 +1,8 @@
 package stats
 
 import (
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/pion/interceptor"
@@ -78,12 +80,9 @@ type recorder struct {
 	maxLastSenderReports          int
 	maxLastReceiverReferenceTimes int
 
-	incomingRTPChan  chan *incomingRTP
-	incomingRTCPChan chan *incomingRTCP
-	outgoingRTPChan  chan *outgoingRTP
-	outgoingRTCPChan chan *outgoingRTCP
-	getStatsChan     chan Stats
-	done             chan struct{}
+	latestStats *internalStats
+	ms          *sync.Mutex // Locks latestStats
+	running     uint32
 }
 
 func newRecorder(ssrc uint32, clockRate float64) *recorder {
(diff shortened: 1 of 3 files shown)


pkg/stats/stats_recorder.go
```diff
@@ -261,38 +263,13 @@ func (r *recorder) recordIncomingRTCP(latestStats internalStats, v *incomingRTCP
...
-
-		case r.getStatsChan <- Stats{
-			InboundRTPStreamStats:        latestStats.InboundRTPStreamStats,
-			OutboundRTPStreamStats:       latestStats.OutboundRTPStreamStats,
-			RemoteInboundRTPStreamStats:  latestStats.RemoteInboundRTPStreamStats,
-			RemoteOutboundRTPStreamStats: latestStats.RemoteOutboundRTPStreamStats,
-		}:
-		}
-	}
+	atomic.StoreUint32(&r.running, 1)
```
sterlingdeng, line 266: What's the purpose of doing this / why do we need to keep state here? Is there a case in which we attach the interceptor and before `Start` is called, the hooks on the interceptors gets called?

sterlingdeng (commented):
+1 to providing a fix to this issue. I ran into the same problem yesterday where the call was blocked because the `Getter` was called after the interceptor closed.

How are you using this interceptor? My use case is logging / reporting metrics whenever a stream is closed, so on the `UnbindLocalStream` and `UnbindRemoteStream` I would make the call using the `Getter` using the provided `ssrc` in the callbacks. There isn't any useful examples, so I had do spend a bit of time to figure out a way to extract the stats out of the stream.

asticode:
> How are you using this interceptor? My use case is logging / reporting metrics whenever a stream is closed, so on the `UnbindLocalStream` and `UnbindRemoteStream` I would make the call using the `Getter` using the provided `ssrc` in the callbacks. There isn't any useful examples, so I had do spend a bit of time to figure out a way to extract the stats out of the stream.

My use case is charting webrtc ingests stats in realtime. I'm using a ticker to request stats on a regular basis to retrieve proper deltas. Unfortunately when I close the PeerConnection I'm having this problem.

pkg/stats/stats_recorder.go
asticode: I'm not really sure there's a case but as there was both `Start` and `Stop` methods, I figured it was important to behave based on the `running` status.

An alternative would indeed be to remove both `Start` and `Stop` methods and update stats whenever packets are coming/going.

atoppi:
I have encountered a very similar issue while closing Peer Connections but did not investigate much about the root issue.

Disabling the stats interceptor or using the code from this PR seems to fix the issue.

Sean-Der:
Oof sorry I missed this @asticode 

I got pulled away on https://github.com/obsproject/obs-studio/pull/7926 and has been all I have been working on since December. I am going to merge this now (and lots of other stuff)

thanks for the patience. I also added you to the Pion org. Hopefully make it easier to merge/fix stuff without being blocked on me :)

Sean-Der (approved):

asticode:
> Oof sorry I missed this @asticode
> 
> I got pulled away on [obsproject/obs-studio#7926](https://github.com/obsproject/obs-studio/pull/7926) and has been all I have been working on since December. I am going to merge this now (and lots of other stuff)
> 
> thanks for the patience. I also added you to the Pion org. Hopefully make it easier to merge/fix stuff without being blocked on me :)

No worries, thanks for merging! And thanks for the invite too! ❤️

Issue #185, opened 2023-05-01
Rare panic when setting TWCC extension


m1k1o:
There is a panic that originates from calling [`SetExtension`](https://github.com/pion/interceptor/blob/3f299e161f970e6cb9213145aeb34bcce975b5e1/pkg/twcc/header_extension_interceptor.go#LL51C1-L51C1) on header that is `nil`.

Unfortunately, it happened only once, I don't know what lead to it and I did not have atached debugger to see more of it.

```go
panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xd46aa5]
    goroutine 209875 [running]:
    github.com/pion/rtp.(*Header).SetExtension(0xc001666f00, 0x1, {0xc002db0a4c, 0x2, 0x2})
    	/go/pkg/mod/github.com/pion/rtp@v1.7.13/packet.go:397 +0x4a5
    github.com/pion/interceptor/pkg/twcc.(*HeaderExtensionInterceptor).BindLocalStream.func1(0xc003054401?, {0xc001679200, 0x4a4, 0x5b4}, 0xc000bade88?)
    	/go/pkg/mod/github.com/pion/interceptor@v0.1.12/pkg/twcc/header_extension_interceptor.go:51 +0x9f
    github.com/pion/interceptor.RTPWriterFunc.Write(0xc000badf7c?, 0xc002d77300?, {0xc001679200?, 0xc000badef8?, 0x89d0cf?}, 0xc000901740?)
    	/go/pkg/mod/github.com/pion/interceptor@v0.1.12/interceptor.go:83 +0x31
    github.com/pion/interceptor/pkg/nack.(*ResponderInterceptor).resendPackets.func1(0xb8ef?)
```

PR #186, opened 2023-05-14
Fix set extension on nil header

m1k1o:
Fixes https://github.com/pion/interceptor/issues/185

I was not sure about if it should fail silently, ignoring setting header and continue or returning error. I opted for the second, so that it is transparent that something is not right when header is nil.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -17,6 +17,7 @@ Jonathan Müller <jonathan@fotokite.com>
 Kevin Caffrey <kcaffrey@gmail.com>
 Maksim Nesterov <msnesterov@avito.ru>
 Mathis Engelbart <mathis.engelbart@gmail.com>
+Miroslav <sedivy.miro@gmail.com>
 Miroslav Šedivý <sedivy.miro@gmail.com>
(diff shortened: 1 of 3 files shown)


AUTHORS.txt
```diff
@@ -17,6 +17,8 @@ Jonathan Müller <jonathan@fotokite.com>
 Kevin Caffrey <kcaffrey@gmail.com>
 Maksim Nesterov <msnesterov@avito.ru>
 Mathis Engelbart <mathis.engelbart@gmail.com>
+Miroslav <sedivy.miro@gmail.com>
```
stv0g, line 20: Do you mind fixing the commit author, so you appear only once in the list?

stv0g (approved):

stv0g:
Oops, I mixed up the windows and merged it. Nevermind. Having you as duplicate author doesnt hurt :D

AUTHORS.txt
m1k1o: I am curious why it generated me twice in a first place. Maybe it did not recognize character `Š` correctly and then repeated it without my last name.

PR #194, opened 2023-06-26
Enable injecting ticker to Sender Report interceptor

kcaffrey:
#### Description

This lets consumers use the Sender Report interceptor in tests that utilize fake clocks. Previously the current time could be injected, but the interceptor would only run according to real-time based on `time.NewTicker`, but now tests can control when the reports are generated.


mengelbart (approved):
LGTM, thanks!

PR #197, opened 2023-07-26
Don't use out-of-order packets to calculate RTPTime for sender reports

cptpcrd:
#### Description

If a sender report is generated immediately after sending an out-of-order packet (which can happen e.g. in an SFU when forwarding media), the timestamp from the last in-order packet should be extrapolated, since the departure time of an out-of-order packet is is not properly correlated with its RTP timestamp.

Leave an option to re-enable the old behavior.

### DIFF
--- a/pkg/report/sender_interceptor.go
+++ b/pkg/report/sender_interceptor.go
@@ -59,6 +59,8 @@ type SenderInterceptor struct {
 	wg        sync.WaitGroup
 	close     chan struct{}
 	started   chan struct{}
+
+	useLatestPacket bool
 }
 
 func (s *SenderInterceptor) isClosed() bool {
@@ -133,7 +135,7 @@ func (s *SenderInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 // BindLocalStream lets you modify any outgoing RTP packets. It is called once for per LocalStream. The returned method
 // will be called once per rtp packet.
 func (s *SenderInterceptor) BindLocalStream(info *interceptor.StreamInfo, writer interceptor.RTPWriter) interceptor.RTPWriter {
-	stream := newSenderStream(info.SSRC, info.ClockRate)
+	stream := newSenderStream(info.SSRC, info.ClockRate, s.useLatestPacket)
 	s.streams.Store(info.SSRC, stream)
 
 	return interceptor.RTPWriterFunc(func(header *rtp.Header, payload []byte, a interceptor.Attributes) (int, error) {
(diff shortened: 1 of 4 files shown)


kcaffrey:
The motivation behind this PR is that we were getting reports of A/V desync from users of our SFU. After some investigation, it seemed like the issue was related to scenarios where any reordering of packets occurred (such as RTX after loss). Depending on how severe the reordering is, it can be the case where the most recently sent packet is actually quite old. This can cause problems with Chrome's "[RTP to NTP estimator](https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:third_party/webrtc/system_wrappers/source/rtp_to_ntp_estimator.cc;l=75;drc=b5cd13bb6d5d157a5fbe3628b2dd1c1e106203c6;bpv=1;bpt=1)". This estimator works by looking at the most recent 20 Sender Reports and running a linear regression to correlate RTP timestamps to NTP timestamps (which are supposed to represent the same instant in time for the values in Sender Reports).

The RTP to NTP estimator (along with a clock offset estimator) is then used to calculate playout timestamps for the respective audio and video tracks in local time by converting the RTP timestamps for the most recently played media frames; if the playout timestamps differ, then either the audio or video jitter buffer may be grown or shrunk to try to keep them in sync.

Before this PR, it was possible for an old packet to be the most recently arrived/sent packet, so the RTP timestamp would be skewed to look older than it should. This would, in turn, impact the RTP estimator for up to 20 seconds (until the bad SR is cycled out), skewing the playout timestamp estimator, and ultimately actually causing A/V desync when Chrome tries to correct desync that doesn't actually exist.

aler9:
Hello,

The RTP-to-NTP analysis you performed is right.

The issue you encountered is related to the special case in which a SFU deliberately routes unordered packets from an origin to clients without performing reordering. While this does minimize latency, it also causes clients to concurrently emit NACKs whenever there are issues in the origin-to-server connection, not in the server-to-client connections, potentially exhausting the SFU bandwidth when there are thousands of clients.

If you instead perform packet reordering and NACK handling on the SFU and recover packets directly on the SFU, clients won't detect losses (unless packets can't be recovered in the origin-to-server connection), won't emit NACKs and won't have NTP/RTP issues.

Therefore this PR is useless if a SFU is implemented in the right way and if the sender interceptor is placed after the reordering mechanism.

adriancable (approved):
This looks good to me. I support it!

adriancable:
@cptpcrd - I support this PR. Can you kindly rebase it on the base branch and then I can merge it.

kcaffrey:
This is now rebased @adriancable. Thanks!

PR #198, opened 2023-07-27
Partial rewrite of TWCC sender

kcaffrey:
The previous implementation of the TWCC sender interceptor had some inconsistencies with libwebrtc. Namely, if there were missing packets between the last packet in the previous feedback interval and the first received packet in the following feedback interval, then those packets were never included as missing in any feedback.

This is an issue because libwebrtc uses data about lost packets from TWCC feedback in their congestion controller. This means that bursts of loss could go unnoticed by libwebrtc, causing the application to possibly send more data than it would if the loss was properly reported.

Another minor difference was that feedback packets with a single packet are in fact valid according to the RFC, but we were only returning feedback with at least two packets. libwebrtc has a check for minimum feedback size, which has now been added to the unit tests.

Just about all of the rewrite here has been ported from libwebrtc code, so it should now match the behavior fairly well.

### DIFF
--- a/pkg/twcc/arrival_time_map.go
+++ b/pkg/twcc/arrival_time_map.go
@@ -1,0 +1,192 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package twcc
+
+const (
+	minCapacity        = 128
+	maxNumberOfPackets = 1 << 15
+)
+
+// packetArrivalTimeMap is adapted from Chrome's implementation of TWCC, and keeps track
+// of the arrival times of packets. It is used by the TWCC interceptor to build feedback
+// packets.
+// See https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:third_party/webrtc/modules/remote_bitrate_estimator/packet_arrival_map.h;drc=b5cd13bb6d5d157a5fbe3628b2dd1c1e106203c6
+type packetArrivalTimeMap struct {
+	// arrivalTimes is a circular buffer, where the packet with sequence number sn is stored
+	// in slot sn % len(arrivalTimes)
+	arrivalTimes []int64
+
+	// The unwrapped sequence numbers for the range of valid sequence numbers in arrivalTimes.
+	// beginSequenceNumber is inclusive, and endSequenceNumber is exclusive.
+	beginSequenceNumber, endSequenceNumber int64
+}
+
+// AddPacket records the fact that the packet with sequence number sequenceNumber arrived
+// at arrivalTime.
+func (m *packetArrivalTimeMap) AddPacket(sequenceNumber int64, arrivalTime int64) {
+	if m.arrivalTimes == nil {
+		// First packet
+		m.reallocate(minCapacity)
+		m.beginSequenceNumber = sequenceNumber
+		m.endSequenceNumber = sequenceNumber + 1
+		m.arrivalTimes[m.index(sequenceNumber)] = arrivalTime
+		return
+	}
+
+	if sequenceNumber >= m.beginSequenceNumber && sequenceNumber < m.endSequenceNumber {
+		// The packet is within the buffer, no need to resize.
+		m.arrivalTimes[m.index(sequenceNumber)] = arrivalTime
+		return
+	}
+
+	if sequenceNumber < m.beginSequenceNumber {
+		// The packet goes before the current buffer. Expand to add packet,
+		// but only if it fits within the maximum number of packets.
+		newSize := int(m.endSequenceNumber - sequenceNumber)
+		if newSize > maxNumberOfPackets {
+			// Don't expand the buffer back for this packet, as it would remove newer received
+			// packets.
+			return
+		}
+		m.adjustToSize(newSize)
+		m.arrivalTimes[m.index(sequenceNumber)] = arrivalTime
+		m.setNotReceived(sequenceNumber+1, m.beginSequenceNumber)
+		m.beginSequenceNumber = sequenceNumber
+		return
+	}
+
+	// The packet goes after the buffer.
+	newEndSequenceNumber := sequenceNumber + 1
+
+	if newEndSequenceNumber >= m.endSequenceNumber+maxNumberOfPackets {
+		// All old packets have to be removed.
+		m.beginSequenceNumber = sequenceNumber
+		m.endSequenceNumber = newEndSequenceNumber
+		m.arrivalTimes[m.index(sequenceNumber)] = arrivalTime
+		return
+	}
+
+	if m.beginSequenceNumber < newEndSequenceNumber-maxNumberOfPackets {
+		// Remove oldest entries.
+		m.beginSequenceNumber = newEndSequenceNumber - maxNumberOfPackets
+	}
+
+	m.adjustToSize(int(newEndSequenceNumber - m.beginSequenceNumber))
+
+	// Packets can be received out of order. If this isn't the next expected packet,
+	// add enough placeholders to fill the gap.
+	m.setNotReceived(m.endSequenceNumber, sequenceNumber)
+	m.endSequenceNumber = newEndSequenceNumber
+	m.arrivalTimes[m.index(sequenceNumber)] = arrivalTime
+}
+
+func (m *packetArrivalTimeMap) setNotReceived(startInclusive, endExclusive int64) {
+	for sn := startInclusive; sn < endExclusive; sn++ {
+		m.arrivalTimes[m.index(sn)] = -1
+	}
+}
+
+// BeginSequenceNumber returns the first valid sequence number in the map.
+func (m *packetArrivalTimeMap) BeginSequenceNumber() int64 {
+	return m.beginSequenceNumber
+}
+
+// EndSequenceNumber returns the first sequence number after the last valid sequence number in the map.
+func (m *packetArrivalTimeMap) EndSequenceNumber() int64 {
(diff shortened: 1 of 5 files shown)


kcaffrey:
I see #202 got merged, but I think this PR would actually replace it and do the same deferred sorting but in O(N) instead of O(N log N). This is because I walk through the sequence numbers in order to build feedback (note that since missing packets have to be added anyway, this adds no overhead). Simply keeping track of the start/end sequence numbers and storing arrival times in a circular buffer lets you do a simple for loop to get the sorted arrival times; no explicit sort needed.

@davidzhao @treyhakanson can you take a look and confirm that my logic above is correct, and that the proper way to resolve this change with #202 is to simply overwrite the changes?

pkg/twcc/twcc.go
```diff
@@ -33,68 +37,118 @@ type Recorder struct {
...
-			return list
-		}
+
+	r.arrivalTimeMap.AddPacket(unwrappedSN, arrivalTime)
+
+	// Limit the range of sequence numbers to send feedback for.
+	if *r.startSequenceNumber < r.arrivalTimeMap.BeginSequenceNumber() {
+		sn := r.arrivalTimeMap.BeginSequenceNumber()
+		r.startSequenceNumber = &sn
+	}
```
boks1971, line 67: Can this check be moved to when building feedback? If an out-of-order comes, it could again get back to some value and again reset here because it is before the beginning sequence number. Wondering if it can be done just when building feedback.

While I do not see any issues in the current implemtation, wondering if the implementation can be done without exposing the `BeginSequenceNumber()`, `EndSequenceNumber()` API from the arrival time map. Those could be a bit risky, i. e. if users do things like calculating size based on those values, etc, it could lead to unintentional errors. Again, nothing of that sort in this PR, but just being cautious. Will leave a few more comments along those lines below.

Thought about this when I was reading `EraseTo` function which resets those numbers (a bunch of other places change them also, but that function caught my eye as begin and end are set to same to indicate empty, but was thinking of chances of users of that arrival map module holding on to stale/outdated begin/end). BTW, did not see use of `EraseTo` API except in tests.

Along those lines would actually be good if this module can eliminate even holding the start sequence. I have not thought about if that can also be rolled into the map (i. e. the map is always source of truth), but maybe worth looking at.

pkg/twcc/twcc.go
```diff
@@ -33,68 +37,118 @@ type Recorder struct {
...
+}
+
+func (r *Recorder) maybeCullOldPackets(sequenceNumber int64, arrivalTime int64) {
+	if r.startSequenceNumber != nil && *r.startSequenceNumber >= r.arrivalTimeMap.EndSequenceNumber() && arrivalTime >= packetWindowMicroseconds {
+		r.arrivalTimeMap.RemoveOldPackets(sequenceNumber, arrivalTime-packetWindowMicroseconds)
```
boks1971, line 72: maybe, this can call `RemoveOldPackets` unconditionally and that API can do the needful. As this is called on every packet, that method can do fast path check as efficiently as possible and return early.

pkg/twcc/twcc.go
```diff
@@ -33,68 +37,118 @@ type Recorder struct {
...
+	}
+	return feedbacks
+}
+
+// maybeBuildFeedbackPacket builds a feedback packet starting from startSN (inclusive) until
+// endSN (exclusive).
+func (r *Recorder) maybeBuildFeedbackPacket(beginSeqNumInclusive, endSeqNumExclusive int64) *feedback {
+	// NOTE: The logic of this method is inspired by the implementation in Chrome.
+	// See https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:third_party/webrtc/modules/remote_bitrate_estimator/remote_estimator_proxy.cc;l=276;drc=b5cd13bb6d5d157a5fbe3628b2dd1c1e106203c6
+	startSN, endSN := r.arrivalTimeMap.Clamp(beginSeqNumInclusive), r.arrivalTimeMap.Clamp(endSeqNumExclusive)
```
boks1971, line 103: maybe call this `startSNInclusive` and `endSNExclusive` to make this explicit?

pkg/twcc/twcc.go
```diff
@@ -33,68 +37,118 @@ type Recorder struct {
...
+	if r.startSequenceNumber == nil {
 		return nil
 	}
 
-	feedback := newFeedback(r.senderSSRC, r.mediaSSRC, r.fbPktCnt)
-	r.fbPktCnt++
-	feedback.setBase(uint16(r.receivedPackets[0].sequenceNumber&0xffff), r.receivedPackets[0].arrivalTime)
+	endSN := r.arrivalTimeMap.EndSequenceNumber()
+	var feedbacks []rtcp.Packet
+	for *r.startSequenceNumber < endSN {
```
boks1971, line 84: maybe, avoid calling `EndSequenceNumber()` by doing something like `for !done{}` and build feedback packet method returns a `bool` to indicate `done`?

pkg/twcc/twcc.go
```diff
@@ -33,68 +37,118 @@ type Recorder struct {
...
 		}
+
+		nextSequenceNumber = seq + 1
 	}
-	r.receivedPackets = []pktInfo{}
-	pkts = append(pkts, feedback.getRTCP())
 
-	return pkts
+	r.startSequenceNumber = &nextSequenceNumber
+	return fb
```
boks1971, line 151: guess it is possible that some packets could be duplicated in feedback. But, I believe it should be fine. Just wanted to make sure that it is fine. For example, will the following send duplicate feedback?
- Packets 11, 12, 13 - feedback sent
- Packets 14, 15, 16, 10 - arrive -> 10 arriving would have set the start sequence number back to 10.
- Next feedback - 10, 11, 12, 13, 14, 15, 16 -> 11, 12 , 13 duplicated, but they should have the same data

boks1971 (approved):
Great stuff @kcaffrey !! LGTM!

But, would be good to get some more 👀 . Will also ask @treyhakanson who recently worked on this.

pkg/twcc/arrival_time_map.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	// The packet goes after the buffer.
+	newEndSequenceNumber := sequenceNumber + 1
+
+	if newEndSequenceNumber >= m.endSequenceNumber+maxNumberOfPackets {
+		// All old packets have to be removed.
+		m.beginSequenceNumber = sequenceNumber
+		m.endSequenceNumber = newEndSequenceNumber
+		m.arrivalTimes[m.index(sequenceNumber)] = arrivalTime
```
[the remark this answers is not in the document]
kcaffrey: Fair point, but I wanted to stay consistent with [libwebrtc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/remote_bitrate_estimator/packet_arrival_map.cc;l=65;bpv=1;bpt=1) here

pkg/twcc/arrival_time_map.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+// EraseTo erases all elements from the beginning of the map until sequenceNumber.
+func (m *packetArrivalTimeMap) EraseTo(sequenceNumber int64) {
+	if sequenceNumber < m.beginSequenceNumber {
+		return
+	}
+	if sequenceNumber >= m.endSequenceNumber {
+		// Erase all.
+		m.beginSequenceNumber = m.endSequenceNumber
```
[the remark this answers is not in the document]
kcaffrey: Same as prior comment: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/remote_bitrate_estimator/packet_arrival_map.cc;l=122;bpv=1;bpt=1

pkg/twcc/twcc.go
kcaffrey: `packetArrivalTimeMap` itself is unexported, so I don't believe users will be able to call those methods. I think you raise some good points, but since they are private APIs, and consistent with libwebrtc, I think I'll leave as-is

pkg/twcc/twcc.go
kcaffrey: Done!

pkg/twcc/twcc.go
kcaffrey: I believe libwebrtc expects duplicated feedback on reorderings: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/remote_bitrate_estimator/remote_estimator_proxy_unittest.cc;l=340;bpv=0;bpt=0. I can't find anything in the [RFC](https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01) that says one way or the other, so defaulting to libwebrtc's behavior seems reasonable here.

pkg/twcc/twcc.go
kcaffrey: I haven't benchmarked this, but I suspect the difference is very minor. In contrast, keeping a direct port of libwebrtc makes it easier to audit that they have the same behavior, so I think even if there is a slight performance hit, it's worth it! It's worth noting that this implementation is O(1) insert (amortized) and O(N) build feedback, which should be much faster than what's on master currently!

pkg/twcc/twcc.go
kcaffrey: I think `EndSequenceNumber()` will almost certainly get inlined, so I think in practice the performance will be the same.

kcaffrey:
Ok, I have reverted #202 and rebased this branch on top of master. I left it as two separate commits to clarify the revert of #202.

PR #202, opened 2023-08-28
Defer packet ordering until building RTCP packet

treyhakanson:
#### Description

The previous implementation did a sorted insert (O(m)) on every call to `Record`. If sorting is deferred until a feedback packet is built, we can record in constant time and build a packet in O(nlogn). Ordering isn't required until building the packet anyway, and deferring nets a minor performance gain (I say minor since its unlikely there are a large number of received packets in the buffer prior to building a feedback packet). If ordering is really needed on record, we could use something like a B-tree to get O(logn) sorted inserts.

#### Reference issue

N/A

### DIFF
--- a/pkg/twcc/twcc.go
+++ b/pkg/twcc/twcc.go
@@ -6,6 +6,7 @@ package twcc
 
 import (
 	"math"
+	"sort"
 
 	"github.com/pion/rtcp"
 )
@@ -44,10 +45,8 @@ func (r *Recorder) Record(mediaSSRC uint32, sequenceNumber uint16, arrivalTime i
 	if sequenceNumber < 0x0fff && (r.lastSequenceNumber&0xffff) > 0xf000 {
 		r.cycles += 1 << 16
 	}
-	r.receivedPackets = insertSorted(r.receivedPackets, pktInfo{
-		sequenceNumber: r.cycles | uint32(sequenceNumber),
-		arrivalTime:    arrivalTime,
-	})
+	pkt := pktInfo{r.cycles | uint32(sequenceNumber), arrivalTime}
+	r.receivedPackets = append(r.receivedPackets, pkt)
 	r.lastSequenceNumber = sequenceNumber
 }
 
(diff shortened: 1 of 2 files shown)


pkg/twcc/twcc.go
```diff
@@ -19,7 +20,7 @@ type pktInfo struct {
 // transport wide congestion control feedback reports as specified in
 // https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01
 type Recorder struct {
-	receivedPackets []pktInfo
+	receivedPackets map[uint32]pktInfo
```
boks1971, line 23: Based on very limited `golang` knowledge, `map`s seem to incur more memory (expected, but making the comment in the sense I have been surprised some times maps growing unexpectedly). Is there a need to use map here? 

Also, this is constructing the slice from map in building feedback packet. That can be avoided I think if we just append and sort that slice directly.

boks1971 (commented):
lgtm @treyhakanson . 

Let me request more :eyes: on this.

pkg/twcc/twcc.go
treyhakanson: @boks1971 thanks for the comment! I used a map here to handle the need to de-dupe packets with the same sequence number (see the previous `insertSorted` method). I could use a slice and de-dupe during iteration when building the feedback packet, but this would require a map/set anyway

boks1971: Thank you @treyhakanson . As these maps should be small, it should be fine. On that one comment, I think re-allocating the `map` after building report would be needed to ensure that map does not grow. Did not see it. Maybe, I missed it in the diff.

pkg/twcc/twcc.go
```diff
@@ -56,50 +55,42 @@ func (r *Recorder) PacketsHeld() int {
...
-			pkts = append(pkts, feedback.getRTCP())
+			rtcpPkts = append(rtcpPkts, feedback.getRTCP())
 			feedback = newFeedback(r.senderSSRC, r.mediaSSRC, r.fbPktCnt)
 			r.fbPktCnt++
 			feedback.addReceived(uint16(pkt.sequenceNumber&0xffff), pkt.arrivalTime)
 		}
 	}
-	r.receivedPackets = []pktInfo{}
-	pkts = append(pkts, feedback.getRTCP())
+	r.receivedPackets = make(map[uint32]pktInfo)
```
boks1971, line 90: Ignore my comment above. See it now.

treyhakanson:
For more context on the motivation behind this change, see this discussion on the pion slack channel: https://gophers.slack.com/archives/CAK2124AG/p1692822880632899

pkg/twcc/twcc.go
mengelbart: You can do the deduplication when building the feedback packet without a map. When iterating over the sorted slice, you remember the previous value and compare it to the next because in the sorted slice, all duplicates will be neighbors. It could be somewhat fiddly, depending on whether you want to insert the first or last occurrence. Currently, we take the last (which would be more complicated), but I think we may want to use the first instead (ordered by arrival time, though, not by the time it was added). That could be added as a criterion to the sort function, then it would be easy again (just check if the current sequence number was already added in the last iteration). However, all this is probably very hypothetical since there will rarely be duplicates anyway.

mengelbart (approved):

pkg/twcc/twcc.go
treyhakanson: Ah you're totally right, not sure why I didn't think of that; thanks for calling out!
Updated to implement this approach and take the packet with the earliest arrival time. I also updated `TestDuplicatePackets` to test that we take the packet with the earliest arrival time even if it isn't the first one we record. Fwiw, I've been testing on LTE/5G networks recently and have noticed a fair bit of duplication so its seems worth handling

mengelbart: Oh, that's good to know, thanks!
And thanks for implementing this change!

PR #207, opened 2023-09-27
TWCC: Fix small receive delta rounding

kcaffrey:
Prior to this change, a sequence of small receive deltas could get erroneously encoded as a series of 0s.

If the actual receive deltas were {200μs, 200μs, ..., 200μs}, what was appearing in the TWCC feedback packet (after marshaling) was {0, 0, ..., 0}. (Note: the added test, without the code fixes, would fail by returning deltas of {200, 200, ..., 200}, but marshalling the feedback packet would truncate all of those to 0).

To fix this, receive deltas are rounded to the closest "delta tick" period (250μs) as opposed to truncating. Additionally, `lastTimestampUS` is moved forward by the rounded delta instead of set to the timestamp of the last packet. This lets rounding errors accumulate so that a series of packets will have the correct sum of receive deltas (within the margin of error of a delta tick).

### DIFF
--- a/pkg/twcc/twcc.go
+++ b/pkg/twcc/twcc.go
@@ -154,10 +154,16 @@ func (f *feedback) getRTCP() *rtcp.TransportLayerCC {
 
 func (f *feedback) addReceived(sequenceNumber uint16, timestampUS int64) bool {
 	deltaUS := timestampUS - f.lastTimestampUS
-	delta250US := deltaUS / 250
+	var delta250US int64
+	if deltaUS >= 0 {
+		delta250US = (deltaUS + rtcp.TypeTCCDeltaScaleFactor/2) / rtcp.TypeTCCDeltaScaleFactor
+	} else {
+		delta250US = (deltaUS - rtcp.TypeTCCDeltaScaleFactor/2) / rtcp.TypeTCCDeltaScaleFactor
+	}
 	if delta250US < math.MinInt16 || delta250US > math.MaxInt16 { // delta doesn't fit into 16 bit, need to create new packet
 		return false
 	}
+	deltaUSRounded := delta250US * rtcp.TypeTCCDeltaScaleFactor
 
 	for ; f.nextSequenceNumber != sequenceNumber; f.nextSequenceNumber++ {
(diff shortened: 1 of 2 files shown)


pkg/twcc/twcc_test.go
```diff
@@ -255,6 +255,53 @@ func Test_feedback(t *testing.T) {
 		}
 	})
 
+	t.Run("add received small deltas", func(t *testing.T) {
+		f := newFeedback(0, 0, 0)
+		base := int64(320 * 1000)
+		deltaUS := int64(200)
+		f.setBase(5, base)
+
+		for i := int64(0); i < 5; i++ {
```
boks1971, line 264: Can probably start the loop from 1? It is re-adding the base which is probably fine. `addReceived` code will cycle through the 16-bit space in the loop here 
```
    for ; f.nextSequenceNumber != sequenceNumber; f.nextSequenceNumber++ {                                                   
        if !f.lastChunk.canAdd(rtcp.TypeTCCPacketNotReceived) {                                                              
            f.chunks = append(f.chunks, f.lastChunk.encode())                                                                
        }                                                                                                                    
        f.lastChunk.add(rtcp.TypeTCCPacketNotReceived)                                                                       
        f.sequenceNumberCount++                                                                                              
    } 
```
which will affect the sequence number count. But, that is not tested here. Can change to loop to start from `1` to avoid that.

pkg/twcc/twcc.go
```diff
@@ -183,9 +188,9 @@ func (f *feedback) addReceived(sequenceNumber uint16, timestampUS int64) bool {
 	f.lastChunk.add(recvDelta)
 	f.deltas = append(f.deltas, &rtcp.RecvDelta{
 		Type:  recvDelta,
-		Delta: deltaUS,
+		Delta: delta250US * 250,
 	})
-	f.lastTimestampUS = timestampUS
+	f.lastTimestampUS += delta250US * 250
```
boks1971, line 193: Existing code was using hard coded `250`, so this PR is not changing that, but would be good to use `rtcp.TypeTCCDeltaScaleFactor`

boks1971 (approved):
lgtm!

kcaffrey:
@boks1971 Thanks for the review, I made those changes.

PR #210, opened 2023-10-16
Fix wrap around on out of order packet

boks1971:
Sequence number cycle calculation was over estimating the cycles if packets came out of order. A sequence like 65535, 0, 65534 would increment the cycle counter twice.

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -3,17 +3,17 @@
 #
 # This file is auto generated, using git to list all individuals contributors.
 # see https://github.com/pion/.goassets/blob/master/scripts/generate-authors.sh for the scripting
- Sean DuBois <sean@siobud.com>
- ziminghua <565209960@qq.com>
 Aaron Boushley <boushley@gmail.com>
 Adam Kiss <masterada@gmail.com>
 adamroach <adam@nostrum.com>
 Aditya Kumar <k.aditya00@gmail.com>
+Adrian Cable <6544927+adriancable@users.noreply.github.com>
 aler9 <46489434+aler9@users.noreply.github.com>
 Antoine Baché <antoine@tenten.app>
 Atsushi Watanabe <atsushi.w@ieee.org>
 Bobby Peck <rpeck@mux.com>
 boks1971 <raja.gobi@tutanota.com>
+cptpcrd <31829097+cptpcrd@users.noreply.github.com>
 David Zhao <david@davidzhao.com>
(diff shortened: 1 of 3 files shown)


cnderrauber (approved):

davidzhao (approved):
lg!

adriancable:
@boks1971 - immediately after the lines of code you've changed is this:

```
			// set missing packets as missing
			for i := stream.lastSeqnum + 1; i != pktHeader.SequenceNumber; i++ {
				stream.delReceived(i)
			}
```

I may be having a real bad code comprehension day, but this doesn't seem right. `pktHeader.SequenceNumber` is the 'non-unwrapped' sequence number, so if for example we have the packet sequence 65533, 65534, 3, 4, 5, ... then when we receive packet with sequence number 3, stream.lastSeqnum will be 65534, pktHeader.SequenceNumber will be 3 and so we won't call delReceived on the missing packets 65535, 1, 2.

Don't we need something like this instead?

```
			for i := uint32(stream.lastSeqnum) + 1; i != uint32(stream.seqnumCycles)<<16 | uint32(stream.lastSeqnum); i++ {
```

boks1971:
> @boks1971 - immediately after the lines of code you've changed is this:
> 
> ```
> // set missing packets as missing
> for i := stream.lastSeqnum + 1; i != pktHeader.SequenceNumber; i++ {
> 	stream.delReceived(i)
> }
> ```
> 
> I may be having a real bad code comprehension day, but this doesn't seem right. `pktHeader.SequenceNumber` is the 'non-unwrapped' sequence number, so if for example we have the packet sequence 65533, 65534, 3, 4, 5, ... then when we receive packet with sequence number 3, stream.lastSeqnum will be 65534, pktHeader.SequenceNumber will be 3 and so the for loop won't execute at all, and we won't call delReceived on the missing packets 65535, 0, 1, 2.
> 
> Don't we need something like this instead?
> 
> ```
> for i := uint32(stream.lastSeqnum) + 1; i != uint32(stream.seqnumCycles)<<16 | uint32(stream.lastSeqnum); i++ {
> ```

It should be fine @adriancable as the types are `uint16`. `i` will be `uint16` and looping over it means it will loop through `65535, 0, 1, 2`. That pattern is used in a lot of places for wrap around handling.

adriancable:
@boks1971 - you are absolutely right of course! Thank you for fixing my bad code comprehension day. :-)

PR #212, opened 2023-10-24
[FlexFec] Adding FlexFec Flex variant encoder implementation

pougetat:
#### Description

- Context:

FlexFec is a packet loss recovery technique through which redundant RTP packets are sent and from which missing RTP packets can potentially be reconstructed.
https://datatracker.ietf.org/doc/html/rfc8627

- Summary:

There are multiple variants of FlexFec and I'm adding an implementation of the "Flex" encoder as part of this PR.
The other variants are not included nor is the decoder included which I hope to add in subsequent PRs, this first one is meant to be the start of a series.
This PR also doesn't add the interceptor on top of the encoder.

I'm labeling it as DRAFT for now until I write some proper unit tests to get us some confidence that it works as expected.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/flexfec/flexfec_encoder.go
+++ b/pkg/flexfec/flexfec_encoder.go
@@ -1,0 +1,184 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+// Package flexfec implements FlexFEC to recover missing RTP packets due to packet loss.
+// https://datatracker.ietf.org/doc/html/rfc8627
+package flexfec
+
+import (
+	"encoding/binary"
+
+	"github.com/pion/interceptor/pkg/flexfec/util"
+	"github.com/pion/rtp"
+)
+
+const (
+	// BaseRTPHeaderSize represents the minium RTP packet header size in bytes.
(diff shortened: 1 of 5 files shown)


Sean-Der:
Hey @pougetat it is great to see you :)

If you can fix those lint errors I would like to merge this. We can add tests/improve other things later.

pkg/flexfec/flexfec_encoder.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return rtp.Packet{
+		Header: rtp.Header{
+			Version:        2,
+			Padding:        false,
+			Extension:      false,
+			Marker:         false,
+			PayloadType:    flex.payloadType,
+			SequenceNumber: flex.baseSN,
+			Timestamp:      54243243,
```
evgmel, line 83: Hi @pougetat. Could you please explain this fixed Timestamp value = 54243243? Is it taken from RFC?

pougetat: It's a leftover hardcoded value. I'll get this fixed soon.

evgmel: Okay. No problem 👍  Thanks for the explanation)

PR #214, opened 2023-11-01
[FlexFec] Adding support for FlexFec RFC-03

pougetat:
##### Context

This PR is part of a series of PRs to provide FlexFec support for pion.
https://datatracker.ietf.org/doc/html/rfc8627

##### Summary

This PR provides the following:

- Adding a FlexFec encoder implementation for https://datatracker.ietf.org/doc/html/draft-ietf-payload-flexible-fec-scheme-03 => This is the implementation that chromium currently uses instead of the latest version of the draft. Providing support for this draft version will enable pion to work with chromium-based browsers.
- Changing the FEC coverage data structure to use 2 uint64 to store the packet masks. This will reduce allocations.
- Adding support for the actual Interceptor component that batches RTP packets together to generate repair packets before sending both types of packets over the network.
- Various fixes to FEC header encoding, validated with chromium.

There are still some issues to be tackled:
- Hardcoded timestamp for the RTP header. Chromium seems to work just fine with this but we need to properly set this value.
- We currently limit ourselves to VP8, FlexFec is meant to be codec agnostic so we can provide support for more codecs.
- We currently hardcode a minimum of 5 media packets protected by 2 media packets, this should be configurable.
- We need to account for missing media packets, indeed media packets could be dropped on the way to being received by pion, and our FEC encoder assumes there are no missing media packets.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/flexfec/encoder_interceptor.go
+++ b/pkg/flexfec/encoder_interceptor.go
@@ -1,0 +1,79 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package flexfec
+
+import (
+	"github.com/pion/interceptor"
+	"github.com/pion/rtp"
+)
+
+// FecInterceptor implements FlexFec.
+type FecInterceptor struct {
+	interceptor.NoOp
+	flexFecEncoder     FlexEncoder
+	packetBuffer       []rtp.Packet
+	minNumMediaPackets uint32
+}
+
+// FecOption can be used to set initial options on Fec encoder interceptors.
+type FecOption func(d *FecInterceptor) error
+
+// FecInterceptorFactory creates new FecInterceptors.
+type FecInterceptorFactory struct {
+	opts []FecOption
+}
+
+// NewFecInterceptor returns a new Fec interceptor factory.
+func NewFecInterceptor(opts ...FecOption) (*FecInterceptorFactory, error) {
+	return &FecInterceptorFactory{opts: opts}, nil
+}
+
+// NewInterceptor constructs a new FecInterceptor.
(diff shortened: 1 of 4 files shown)


pkg/flexfec/encoder_interceptor.go
```diff
@@ -0,0 +1,82 @@
...
+
+type FECInterceptorFactory struct {
+	opts []FECEncoderOption
+}
+
+func streamSupportFec(info *interceptor.StreamInfo) bool {
+	// Need to check stream info to see if we should be sending FEC packets
+	if info.MimeType == "video/VP8" {
+		return true
+	}
```
pougetat, line 28: We should support more than just VP8.

pkg/flexfec/encoder_interceptor.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (r *FECInterceptorFactory) NewInterceptor(_ string) (interceptor.Interceptor, error) {
+	// Hardcoded for now:
+	// Min num media packets to encode FEC -> 5
+	// Min num fec packets -> 1
+
+	interceptor := &FecInterceptor{
+		packetBuffer:       make([]rtp.Packet, 0),
+		minNumMediaPackets: 5,
```
pougetat, line 43: This should be configurable

SerialVelocity:
Hey @pougetat, does this only work for Opus (since it has `inbandfec=1`) or will it work for VP8, etc as well?
I saw that the (Chrome) SDP has:
```
a=rtpmap:49 flexfec-03/90000
a=rtcp-fb:49 goog-remb
a=rtcp-fb:49 transport-cc
a=fmtp:49 repair-window=10000000
```
I'm assuming that means that flexfec-03 has to be sent over 49 instead of something like 96 or am I misunderstanding something?

Is this interceptor plug and play or do I need to set up extra lines in the SDP/RTCPFeedback/etc?

PR #224, opened 2024-01-24
Reset exponentialMovingAverage added

alexpokotilo:
Reset exponentialMovingAverage if last decrease was 1 minute ago or If R_hat(i) increases above three standard deviations of the average max bitrate

#### Description
According to https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02#section-4.4 we should reset the average

` If R_hat(i) increases above three standard deviations of the average
   max bitrate, we assume that the current congestion level has changed,
   at which point we reset the average max bitrate and go back to the
   multiplicative increase state.`

I also added average reset in a minute after last update in decrease state.
If we don't reset this average we could eventually encounter with the case when  current receive rate is in average+-3*deviation
of last decrease average. But this average for invented to stay in increment state for short period of time but not forever.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/gcc/rate_controller.go
+++ b/pkg/gcc/rate_controller.go
@@ -34,20 +34,35 @@ type rateController struct {
 }
 
 type exponentialMovingAverage struct {
+	init         bool
 	average      float64
 	variance     float64
 	stdDeviation float64
+	lastUpdate   time.Time
+}
+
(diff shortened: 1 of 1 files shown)


alexpokotilo:
@mengelbart could you please review this request?

alexpokotilo:
new request submitted containing this fix

PR #225, opened 2024-01-24
Don't double update uncertainty

alexpokotilo:
Standard says we need to use either 3*sqrt(var_v_hat) or z(i)

#### Description
Here is quote from

`
If z(i) > 3*sqrt(var_v_hat) the filter is updated with 3*sqrt(var_v_hat) rather than z(i)
`

rather means we need to update uncertainty with 3*sqrt(var_v_hat) and don't add z(i) this time

#### Reference issue
Fixes #...


alexpokotilo:
@mengelbart could you please review  ?

alexpokotilo:
new request submitted containing this fix

PR #226, opened 2024-01-25
Bandwidth estimator improvements

alexpokotilo:
#### Description
calculate fmax for Kalman filter as stated in standard.
Kalman filter stats returning sensitive estimate and we don't need to multiply it with 60.
Kalman gain should not fall lower than minimal value or filter become very insensitive to z(i).
Set Kalman chi to chi to make it more sensitive to input and react faster.
Fix exponentialMovingAverage variance calculation for rate controller.
Reset latestDecreaseRate for rate controller as standard states.
Reset latestDecreaseRate if last decrease was more than a minute ago. In this case latestDecreaseRate in inaccurate in anyway.
Merge https://github.com/pion/interceptor/pull/221 to prevent bandwidth decrease in increase state in last decrease was recently.
Removed not used 'type estimatorFunc'  

There are not fixed tests exists. If my work appreciated and tend to be accepted I will fix all tests as well. I need to know whether anybody want to review these changes.
Right now I was able to fix everything I was able to find.
@mengelbart, @Sean-Der  please take a look or point me to somebody I can review this work.
I spend several weeks trying to understand and fix this BWE. I'm not sure I understand everything but enough to pay attention to this effort. Please let me know if gcc in Pion is important, supported component   

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/gcc/adaptive_threshold.go
+++ b/pkg/gcc/adaptive_threshold.go
@@ -8,10 +8,6 @@ import (
 	"time"
 )
 
-const (
-	maxDeltas = 60
-)
-
 type adaptiveThresholdOption func(*adaptiveThreshold)
 
 func setInitialThreshold(t time.Duration) adaptiveThresholdOption {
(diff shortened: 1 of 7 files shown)


alexpokotilo:
Fixed tests and golint checks

alexpokotilo:
@mengelbart, @Sean-Der  tests are fixed. Could you please re-run tests to start review ?

PR #227, opened 2024-01-29
Keep only 500ms buffer items for leaky pacer

alexpokotilo:
Don't allow LeakyBucketPacer to grow forever in case if client produces more bitrate than BWE proposed

#### Description
Current implementation of leaky pacer expects client to strictly honor bitrate proposed by BWE. But what if client reacts slowly or has an issue in bitrate logic. Leaky pacer show keep 500ms of items and drop older items.
#### Reference issue
Fixes #...

### DIFF
--- a/pkg/gcc/leaky_bucket_pacer.go
+++ b/pkg/gcc/leaky_bucket_pacer.go
@@ -7,6 +7,7 @@ import (
 	"container/list"
 	"errors"
 	"sync"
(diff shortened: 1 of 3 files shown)


alexpokotilo:
@mengelbart please review

alexpokotilo:
Will proceed work on my branch. Let me know if somebody is interested

PR #254, opened 2024-05-02
fix: Some bad math with the JitterBuffer sequence tracking

### LINKED ISSUE #253: Jitter Buffer error
### Your environment.
  - Version: *v0.1.29 and main branch*

### What did you do?
I push and pop packets to the Jitter Buffer.

### What did you expect?
If I added N packets then I can get N packets.

### What happened?
If I pushed packet with a SequenceNumber equal MaxUint16, then I won't be able to pop it. Packets with this SequenceNumber are possible.

I think the problem is in the following [code](https://github.com/pion/interceptor/blob/142f17f90b527434b62ce2fd92676e6e262cd807/pkg/jitterbuffer/jitter_buffer.go#L218) and similar. 

@thatsnotright Have a look, please.

thatsnotright:
#### Description
I'm not sure what I was thinking, uint16 should wrap itself correctly.

#### Reference issue
Fixes #253

### DIFF
--- a/pkg/jitterbuffer/jitter_buffer.go
+++ b/pkg/jitterbuffer/jitter_buffer.go
@@ -7,7 +7,6 @@ package jitterbuffer
 
 import (
 	"errors"
-	"math"
 	"sync"
 
 	"github.com/pion/rtp"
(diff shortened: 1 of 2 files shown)


thatsnotright:
@Sean-Der Can you let me know if this PR is good to go? Thanks!

Sean-Der (approved):

PR #257, opened 2024-06-17
Unmarshal only actually read bytes

### LINKED ISSUE #255: The Jitter buffer interceptor doesn't respect the actual packet length
### Your environment.
  - Version: 72c0be86db35deb9b0a7361063314923a3d9a68c

### What did you do?
I'm trying to use jitter buffer interceptor.

### What did you expect?
I expect to receive RTP packets as usual, but in correct order.

### What happened?
I receive invalid RTP packets. The payload length is more than expected. The payload is filled by zeroes.

### The reason
The interceptor tries to unmarshal the whole buffer and doesn't respect the actual packet length.
https://github.com/pion/interceptor/blob/142f17f90b527434b62ce2fd92676e6e262cd807/pkg/jitterbuffer/receiver_interceptor.go#L77

It looks like we should replace `packet.Unmarshal(buf)` to `packet.Unmarshal(buf[:n])`

braams:
#### Description
The interceptor tried to unmarshal the whole buffer and didn't respect the actual packet length.
#### Reference issue
Fixes #255

### DIFF
--- a/pkg/jitterbuffer/receiver_interceptor.go
+++ b/pkg/jitterbuffer/receiver_interceptor.go
@@ -74,7 +74,7 @@ func (i *ReceiverInterceptor) BindRemoteStream(_ *interceptor.StreamInfo, reader
 			return n, attr, err
 		}
 		packet := &rtp.Packet{}
-		if err := packet.Unmarshal(buf); err != nil {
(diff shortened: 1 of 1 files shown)


thatsnotright:
Let me add some tests around this, unless you wish to, and then we can get this merged.

PR #258, opened 2024-06-17
Shift the expected sequence number if a packet was lost

### LINKED ISSUE #256: The Jitter buffer iterceptor doesn't handle packet loss
### Your environment.
  - Version: 72c0be86db35deb9b0a7361063314923a3d9a68c

### What did you do?
I'm trying to use jitter buffer interceptor.

### What did you expect?
I expect to get an error on read if the current packet was lost. And I expect to get the next packet on the next read.

### What happened?
If even one packet was lost, all subsequent reads will return an error.

### The reason
The interceptor does not handle the situation with the absence of the next packet and does not shift the expected packet number
https://github.com/pion/interceptor/blob/142f17f90b527434b62ce2fd92676e6e262cd807/pkg/jitterbuffer/receiver_interceptor.go#L73
It looks like we should add:
```
				if errors.Is(err, ErrNotFound) {
					i.buffer.SetPlayoutHead(i.buffer.PlayoutHead() + 1)
				}
```

braams:
#### Description
Handle the absence of the next packet
#### Reference issue
Fixes #256

### DIFF
--- a/pkg/jitterbuffer/jitter_buffer.go
+++ b/pkg/jitterbuffer/jitter_buffer.go
@@ -212,9 +212,10 @@ func (jb *JitterBuffer) Pop() (*rtp.Packet, error) {
 	if err != nil {
 		jb.stats.underflowCount++
 		jb.emit(BufferUnderflow)
+		jb.playoutHead++
 		return nil, err
 	}
-	jb.playoutHead = (jb.playoutHead + 1)
+	jb.playoutHead++
 	jb.updateState()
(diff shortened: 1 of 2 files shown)


thatsnotright:
@braams I don't agree with this PR. The intent is that the jitter buffer is agnostic to how packet loss is handled/detected in the case that it can be recovered externally. That means it is up to the caller of the jitter buffer to decide to skip the packet by setting the playback head and not the jitter buffer internals.

PR #285, opened 2024-10-29
Increment rtx rtp packet sequence number only when trasmitted

oto313:
#### Fix for wrong sequence numbers
Currently the nack responder ([responder_interceptor.go](https://github.com/pion/interceptor/blob/master/pkg/nack/responder_interceptor.go)) generates RTX RTP packet which has not right sequence numbers. It does not increment the sequence number by one according to RFC [rfc4588](https://www.rfc-editor.org/rfc/rfc4588)

> For either multiplexing scheme, the sequence number has the standard definition; i.e., it MUST be one higher than the sequence number of the preceding packet sent in the retransmission stream.

I am proposing fix for it and appreciate any feedback as this is my first pull request in GO lang.

### DIFF
--- a/internal/rtpbuffer/packet_factory.go
+++ b/internal/rtpbuffer/packet_factory.go
@@ -17,7 +17,6 @@ const rtxSsrcByteLength = 2
 // The NoOpPacketFactory doesn't copy packets, while the RetainablePacket will take a copy before adding
 type PacketFactory interface {
 	NewPacket(header *rtp.Header, payload []byte, rtxSsrc uint32, rtxPayloadType uint8) (*RetainablePacket, error)
-	FillSequenceNumber(packet *RetainablePacket)
 }
 
 // PacketFactoryCopy is PacketFactory that takes a copy of packets when added to the RTPBuffer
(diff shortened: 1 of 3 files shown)


Sean-Der:
Nice find @oto313! How did you find this, I am really impressed! If you run `go test -v` locally you have some build/test failures. 

* Maybe we make this part of `Get`? Auto-increments (if needed)
* We should write some tests that it does the auto increment as expected

oto313:
Hi @Sean-Der. Thanks. I was investigating why retransmission not worked in my use case and I was investigating it with wireshark and I compared retransmission packets with google webrtc implementation.

> If you run go test -v locally you have some build/test failures.
Sorry the build/tests are fixed

> Maybe we make this part of Get? Auto-increments (if needed)
part of which `Get`?

oto313:
Hi @Sean-Der, so now i fixed the tests. Hopefully now we can merge it. Thanks

internal/rtpbuffer/packet_factory.go
```diff
@@ -94,8 +94,6 @@ func (m *PacketFactoryCopy) NewPacket(header *rtp.Header, payload []byte, rtxSsr
 		p.header.SSRC = rtxSsrc
 		// Rewrite the payload type.
 		p.header.PayloadType = rtxPayloadType
-		// Rewrite the sequence number.
-		p.header.SequenceNumber = m.rtxSequencer.NextSequenceNumber()
```
aalekseevx, line 98: `m.rtxSequencer` seems unused now, maybe delete it?

PR #287, opened 2024-11-12
gcc: fix missing if-else in kalman

### LINKED ISSUE #286: GCC: Maybe bug in kalman filter
### Your environment.
  - Version: latest

### What happened?

The bug lies in this section of code in the Kalman filter.. 
https://github.com/pion/interceptor/blob/master/pkg/gcc/kalman.go#L81-L84
I think it should be an `if else` statement, as this is what I'm seeing in another implementation https://gitlab.freedesktop.org/gstreamer/gst-plugins-rs/-/blob/main/net/rtp/src/gcc/imp/kalman_estimator.rs?ref_type=heads#L53-59
and how I interpret it from the RFC

> If z(i) > 3*sqrt(var_v_hat) the filter is updated with 3*sqrt(var_v_hat) rather than z(i).

In https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02#section-5.3

It would be great to have another pair of eyes on it

sterlingdeng:
#### Description
Fixes a deviation from the GCC rfc spec.

#### Reference issue
Fixes #286

### DIFF
--- a/pkg/gcc/kalman.go
+++ b/pkg/gcc/kalman.go
@@ -80,8 +80,9 @@ func (k *kalman) updateEstimate(measurement time.Duration) time.Duration {
 		root3 := 3 * root
 		if zms > root3 {
 			k.measurementUncertainty = math.Max(alpha*k.measurementUncertainty+(1-alpha)*root3*root3, 1)
+		} else {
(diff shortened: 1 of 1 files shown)


sterlingdeng:
I'm not sure what's going on with the linter - it doesn't seem to mention the offenders. I tried to fix some locations where newline was added but the linter still isn't passing.

mengelbart (approved):

PR #291, opened 2024-11-24
Fixes bug in group accumulator

### LINKED ISSUE #274: A question about arrival_group_accumulator.go
```
func (a *arrivalGroupAccumulator) run(in <-chan []cc.Acknowledgment, agWriter func(arrivalGroup)) {
	init := false
	group := arrivalGroup{}
	for acks := range in {
		for _, next := range acks {
			if !init {
				group.add(next)
				init = true
				continue
			}
			if next.Arrival.Before(group.arrival) {
				// ignore out of order arrivals
				continue
			}
			if next.Departure.After(group.departure) {
				if interDepartureTimePkt(group, next) <= a.interDepartureThreshold {
					group.add(next)
					continue
				}

				if interArrivalTimePkt(group, next) <= a.interArrivalThreshold &&
					interGroupDelayVariationPkt(group, next) < a.interGroupDelayVariationTreshold {
					group.add(next)
					continue
				}
				// println("Arrival len", len(group.packets), group.packets[len(group.packets)-1].Departure.Sub(group.packets[0].Departure).Milliseconds())
				agWriter(group)
				group = arrivalGroup{}
				group.add(next)
			}
		}
	}
}
```
The above code is excerpted from lines 23 to 55 of arrival_group_accumulator.go. After each new pkt is added, the group's departure_time is updated to the new packet's departure_time. In other words, as long as the time difference between the departure_time of the newly added packet and the previous packet is less than the threshold, it will be added to the group. How can we ensure that the time difference between the first packet and the last packet is less than 5 ms (the threshold)?

@thatsnotright

sterlingdeng:
#### Description

See referenced issue for background

#### Reference issue
Fixes #274

### DIFF
--- a/pkg/gcc/arrival_group.go
+++ b/pkg/gcc/arrival_group.go
@@ -16,10 +16,17 @@ type arrivalGroup struct {
 	arrival   time.Time
 }
 
+func newArrivalGroup(a cc.Acknowledgment) arrivalGroup {
+	return arrivalGroup{
+		packets:   []cc.Acknowledgment{a},
+		departure: a.Departure,
+		arrival:   a.Arrival,
+	}
+}
+
 func (g *arrivalGroup) add(a cc.Acknowledgment) {
 	g.packets = append(g.packets, a)
 	g.arrival = a.Arrival
-	g.departure = a.Departure
 }
 
 func (g arrivalGroup) String() string {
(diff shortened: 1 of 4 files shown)


Sean-Der (approved):

Sean-Der:
@sterlingdeng thank you for going so deep on this code. I really appreciate it!

Mind squashing your commit into one and doing a commit message with Subject+Body (just describing what you did)

sterlingdeng:
@mengelbart FYI since you were the original author..

sterlingdeng:
@Sean-Der No problem, glad to help.

Regarding CI: Not sure why it's failing now.. perhaps flakiness?

PR #294, opened 2024-12-16
Fix nack buffer RTX issue

kcaffrey:
There was previously an issue with the nack buffer that would occur when an uploader was experiencing packet loss (or otherwise had reordering occur). `lastAdded` was getting set for every packet added to the buffer, but packets could get added out of order, causing `lastAdded` to go backwards. The impact was that packets at the end of the buffer were sometimes considered stale and cleared out. Consider the following sequence of packets being added to the buffer:

1, 2, 4, 3, 5, ...

When 4 is added, `lastAdded` would correctly be 4, but then when 3 arrived, `lastAdded` would go backwards to 3. This would cause two related problems: trying to get 4 from the buffer would fail (`uint16(3 - 4) > Uint16SizeHalf`), and then when we add 5, we would treat 4 as a gap between `lastAdded` (3) and `seq` (5) and thus release 4 from the buffer, such that even after 5 was added, we would still not be able to retrieve 4 from the buffer.

This PR modifies the logic to only move `lastAdded` forward if `diff` is "positive" (less than `Uint16SizeHalf`), meaning the new packet is newer than any packet received previously.

### DIFF
--- a/internal/rtpbuffer/rtpbuffer.go
+++ b/internal/rtpbuffer/rtpbuffer.go
@@ -17,10 +17,10 @@ const (
 
 // RTPBuffer stores RTP packets and allows custom logic around the lifetime of them via the PacketFactory
 type RTPBuffer struct {
-	packets   []*RetainablePacket
-	size      uint16
-	lastAdded uint16
-	started   bool
+	packets      []*RetainablePacket
+	size         uint16
+	highestAdded uint16
+	started      bool
 }
 
 // NewRTPBuffer constructs a new RTPBuffer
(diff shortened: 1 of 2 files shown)


internal/rtpbuffer/rtpbuffer.go
```diff
@@ -67,6 +67,7 @@ func (r *RTPBuffer) Add(packet *RetainablePacket) {
 			}
 			r.packets[idx] = nil
 		}
+		r.lastAdded = seq
```
boks1971, line 70: maybe rename to something like `newest` or `highestAdded` or something like that.

kcaffrey: Done

PR #300, opened 2025-01-16
WIP: Add interceptor to aggregate CCFB reports

mengelbart:
#### Description

This interceptor keeps track of outgoing RTP packets and reads incoming CCFB (RFC8888) RTCP packets. For each incoming CCFB packet, it creates a report of the included acknowledgments and stores it in the interceptor attributes. Applications reading RTCP packets from it can read the report, e.g., to perform bandwidth estimation.

This is a PoC to improve bandwidth estimation in Pion. The current GCC implementation is hard to use and test. Giving applications access to the feedback reports could simplify and decouple the actual bandwidth estimation from the interceptors.

#### TODO

- [ ] Add unit tests
- [ ] Implement TWCC converter (currently, only RFC 8888 is supported)
- [ ] Document the semantics of the generated reports (currently, it generates a report for every packet included in the feedback packet, including the sequence number, size, departure timestamp, status (received or not), arrival timestamp, and ECN bits.


Sean-Der:
This is so exciting! I am glad to see you back @mengelbart :) 

@jech and @joeturki are getting deep into this world if you are looking for people to work with/quick reviews

jech:
Could you please give some background?  Why not simply provide a function that parses the reports?

To my untrained eyes, it looks like your interceptor aims to allow a client to do:

```go
_, attr, _ := track.ReadRTCP()
data := attr.Get(CCFBAttributesKey)
```

Without an interceptor, the same code would look like this:
```go
p, _, _ := track.ReadRTCP()
var data CCFBReportPacket
data.Unmarshal(p.Data)
```

Now, perhaps I'm biased against interceptors (which are causing me a fair amount of suffering), but the second code looks clearer to me.  What am I missing?

mengelbart:
I tried to make all of the new linters happy, but I added `nolint` in a few places...

pkg/ccfb/interceptor.go
```diff
@@ -0,0 +1,252 @@
...
+// nolint
+func ccfbConverterFactory(f func(ts time.Time, feedback *rtcp.CCFeedbackReport) (time.Time, map[uint32][]acknowledgement)) Option {
+	return func(i *Interceptor) error {
+		i.convertCCFB = f
+
+		return nil
+	}
+}
+
+func twccConverterFactory(f func(feedback *rtcp.TransportLayerCC) (time.Time, map[uint32][]acknowledgement)) Option {
```
3DRX, line 76: Looks like twcc is supported too (not just rfc8888), should we update the description and title of this PR?

3DRX:
Hi @mengelbart what a cool PR! I'm curious to learn what's the relationship between this interceptor, the pacer, and bandwidth estimator GCC will be like by your design. For users using this interceptor, they need to make sure it goes after the pacer right? Otherwise the captured send time will be incorrect. I also take a brief look at [the GCC refactor](https://github.com/pion/interceptor/pull/306), and it seems like pacing is not included in it.

pkg/ccfb/interceptor.go
```diff
@@ -0,0 +1,252 @@
...
+			switch fb := pkt.(type) {
+			case *rtcp.CCFeedbackReport:
+				reportDeparture, reportLists = i.convertCCFB(now, fb)
+			case *rtcp.TransportLayerCC:
+				reportDeparture, reportLists = i.convertTWCC(fb)
+			default:
+			}
+			ssrcToPrl := map[uint32][]PacketReport{}
+			for ssrc, reportList := range reportLists {
+				prl := i.ssrcToHistory[ssrc].getReportForAck(reportList)
```
JoeTurki, line 235: Correct me if I'm wrong, but could this panic if there's no history?
It also seems like it could be abused to trigger a panic for an SSRC that doesn’t have any history yet (though I’m not sure if that’s actually possible).
Additionally, is there a potential race condition here with BindLocalStream?

JoeTurki (approved):
looks good to me, thank you

pkg/ccfb/interceptor.go
mengelbart: Good points. I added a check and skip for unknown SSRCs. I also made the lock an RWLock so we can concurrently read the history map for sending RTP packets and receive RTCP packets. And I added a lock around the missing header extension error case in the RTPWriter.

JoeTurki: Thanks!

PR #301, opened 2025-01-16
Move NTP function to ntp package and add ToTime32

mengelbart:
Moves all NTP-related functions to the NTP package and adds a function to recover a time.Time from a uint32 encoded NTP timestamp. This allows reusing code for CCFB handling (and testing).

### DIFF
--- a/internal/ntp/ntp.go
+++ b/internal/ntp/ntp.go
@@ -20,6 +20,11 @@ func ToNTP(t time.Time) uint64 {
 	return uint64(integerPart)<<32 | uint64(fractionalPart)
 }
 
+// ToNTP32 converts a time.Time object to a uint32 NTP timestamp
+func ToNTP32(t time.Time) uint32 {
+	return uint32(ToNTP(t) >> 16)
+}
+
 // ToTime converts a uint64 NTP timestamps to a time.Time object
 func ToTime(t uint64) time.Time {
 	seconds := (t & 0xFFFFFFFF00000000) >> 32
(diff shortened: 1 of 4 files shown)


internal/ntp/ntp.go
```diff
@@ -20,6 +20,11 @@ func ToNTP(t time.Time) uint64 {
 	return uint64(integerPart)<<32 | uint64(fractionalPart)
 }
 
+// ToNTP32 converts a time.Time object to a uint32 NTP timestamp
+func ToNTP32(t time.Time) uint32 {
+	return uint32((ToNTP(t) >> 16) & 0xFFFFFFFF)
```
joeturki, line 25: redundant masking?

joeturki (approved):

internal/ntp/ntp.go
mengelbart: good point.

mengelbart: fixed

internal/ntp/ntp.go
```diff
@@ -20,6 +20,11 @@ func ToNTP(t time.Time) uint64 {
 	return uint64(integerPart)<<32 | uint64(fractionalPart)
 }
 
+// ToNTP32 converts a time.Time object to a uint32 NTP timestamp
+func ToNTP32(t time.Time) uint32 {
+	return uint32((ToNTP(t) >> 16))
```
aalekseevx, line 25: `return uint32(ToNTP(t) >> 16)` should work?

aalekseevx (approved):

Issue #299, opened 2025-01-13
Congestion control estimation grows beyond what's reasonable


jech:
Hi,

I'm trying to experiment with Pion's GCC estimator in Galene.

I've found that even though I'm sending a maximum of 784kbit/s, the estimator's estimate keeps growing in the absence of congestion: the estimate grew up to 4Mbit/s before I stopped it.

This is of course not reasonable.  A stream of 784kbit/s indicates that the link remains uncongested at 784kbit/s, there is nothing to indicate how it will behave at a rate of 4Mbit/s.

nils-ohlmeier:
If I'm not mistaken the increasing bandwidth estimates are suppose to be caused by the implementation sending bandwidth probing data intermixed with the actual 786kbit payload. If you look closely you should be able to find this probing data as either padding packets or "useless" retransmissions, e.g. on the RTX track, of already received payload packets.

aalekseevx:
Unfortunately, it's a known issue with the current implementation – I will make a reproduction in bwe-test repo: https://github.com/pion/bwe-test/issues/54

Issue #271, opened 2024-08-09
GCC: RateController does not use the internal state


N1cOs:
### Your environment.
  - Version: v0.1.30

### What did you do?
RateController on the delay stats update doesn't use the internals state. Should we make a transition from the internal state, not from the current one (which is always increase actually)? Snippet from the [rate_controller.go ](https://github.com/pion/interceptor/blob/v0.1.30/pkg/gcc/rate_controller.go):
```go
func (c *rateController) onDelayStats(ds DelayStats) {
	now := time.Now()

	if !c.init {
		c.delayStats = ds
		c.delayStats.State = stateIncrease
		c.init = true
		return
	}
	c.delayStats = ds
	// Transition from current state?
	c.delayStats.State = c.delayStats.State.transition(ds.Usage)

	if c.delayStats.State == stateHold {
		return
	}

	// ... more logic
}
```
### What did you expect?
Using the internal state (smth like this):
```go
func (c *rateController) onDelayStats(ds DelayStats) {
	now := time.Now()

	if !c.init {
		c.delayStats = ds
		c.delayStats.State = stateIncrease
		c.init = true
		return
	}
	
	prev := c.delayStats
	c.delayStats = ds
	c.delayStats.State = prev.State.transition(ds.Usage)

	if c.delayStats.State == stateHold {
		return
	}

	// ... more logic
}
```
### What happened?

sterlingdeng:
I noticed this same issue when doing a port of GCC for QUIC and was looking at different implementations.

Issue #260, opened 2024-07-04
Shifted arrival times when some packets in TransportLayerCC have expired


anders-avos:
### Your environment.
  - Version: master
  - Browser: N/A

### What did you do?
When `TransportLayerCC` packets are parsed in [feedback_adapter.go](https://github.com/pion/interceptor/blob/f42ed2b96855d5edfea5e7189b43ea3b8b2d8569/internal/cc/feedback_adapter.go#L142) the [refTime and deltaIndex](https://github.com/pion/interceptor/blob/f42ed2b96855d5edfea5e7189b43ea3b8b2d8569/internal/cc/feedback_adapter.go#L128) are only updated for packets found in history. If only some packets in the report are missing from history this leads to subsequent packets getting the wrong arrival time. See test below.

### What did you expect?
The arrival time for packets still in history should be correct even if some earlier packets in the report expired.

```go
func TestRecvDeltaOutOfSync(t *testing.T) {
	adapter := NewFeedbackAdapter()

	t0 := time.Time{}
	headers := []rtp.Header{}
	// passes when packet count is <= 250
	for i := uint16(0); i < 251; i++ {
		pkt := getPacketWithTransportCCExt(t, i)
		headers = append(headers, pkt.Header)
		assert.NoError(t, adapter.OnSent(t0, &pkt.Header, 1200, interceptor.Attributes{TwccExtensionAttributesKey: hdrExtID}))
	}

	results, err := adapter.OnTransportCCFeedback(t0, &rtcp.TransportLayerCC{
		Header:             rtcp.Header{},
		SenderSSRC:         0,
		MediaSSRC:          0,
		BaseSequenceNumber: 0,
		PacketStatusCount:  22,
		ReferenceTime:      0,
		FbPktCount:         0,
		PacketChunks: []rtcp.PacketStatusChunk{
			&rtcp.StatusVectorChunk{
				PacketStatusChunk: nil,
				Type:              rtcp.TypeTCCStatusVectorChunk,
				SymbolSize:        rtcp.TypeTCCSymbolSizeTwoBit,
				SymbolList: []uint16{
					rtcp.TypeTCCPacketReceivedSmallDelta,
					rtcp.TypeTCCPacketReceivedSmallDelta,
				},
			},
		},
		RecvDeltas: []*rtcp.RecvDelta{
			{
				Type:  rtcp.TypeTCCPacketReceivedSmallDelta,
				Delta: 3,
			},
			{
[...]

PR #306, opened 2025-01-20
GCC refactoring

mengelbart:
#### Description

This PR refactors the GCC implementation using the new API introduced in #300. Currently, users are responsible for creating a `SendSideController`, reading CCFB reports from the attributes, and passing them to the `SendSideController`. This does not include pacing, which needs to be handled separately. Since the new controller does not need to read or pace outgoing packets, we no longer need the cumbersome callback API to get a pointer to the bandwidth estimator.

This refactoring also cleans up the architecture of the bandwidth estimator. It no longer uses the weird pipelining structure and no longer uses any concurrency. This should also fix/close some of the open issues and PRs like #271, #299, #260, #221, #296.

This should not be merged before #300 is merged. A follow-up PR to this one will clean up the old `pkg/gcc` directory to use this version.


mengelbart:
This PR is basically a rewrite of the bandwidth estimator without using any interceptor-specific concepts. Since it does not add an interceptor and can be used without interceptors, what do you think of moving the package into a separate repository which can then also hold alternative bandwidth estimator implementations? @Sean-Der @aalekseevx @JoeTurki

aalekseevx:
@mengelbart, I like this idea. I think it would be convenient to keep the BWE implementation and tests in the same place. Maybe we can use the existing pion/bwe-test repository for this and rename it to pion/bwe?

JoeTurki:
Hello, I agree with @aalekseevx, let's see what Sean see.

mengelbart:
pion/bwe sounds good. Unfortunately, I never got to really finish the test suite there, but it might be a good opportunity to finally straighten that out.

mengelbart:
I created https://github.com/pion/bwe-test/pull/61

PR #308, opened 2025-01-27
Make refactored BWE compatible with the old API

mengelbart:
#### Description
This makes the refactored BWE implementation compatible with the existing `cc.Interceptor` by implementing the `cc.BandwidthEstimator` interface and wrapping the new `ccfb.Interceptor` in the old `cc.Interceptor`. It does not (yet) implement pacing because that would cause nested pacing with the old implementation. Instead of merging this PR, I would actually prefer to remove the old API completely (by removing the `gcc` and `cc` packages).


JoeTurki:
Again I'm sorry about the conflicts, we had to ship the lint upgrades, Let me know, And I'll fix them for this branch :)

PR #310, opened 2025-01-31
NACK responder: bypass auxiliary SSRCs

aalekseevx:
#### Description

The current implementation of RTX in the NACK interceptor implies that there may be packets with different SSRCs in a single RTP stream (stream in terms of the interceptor interface), e.g., with `SSRCRetransmission` or (more importantly for this PR) `SSRCForwardErrorCorrection`. These packets should not be stored in the buffer or used for responding to NACKs.

I suggest skipping such packets in the responder interceptor. This might be a breaking change for some exotic use cases, but I can't think of any sensible scenario where the NACK interceptor is used to resend packets with a non-main SSRC. Therefore, I believe this is a reasonable default.


nils-ohlmeier (approved):

aalekseevx:
Thank you @nils-ohlmeier :)

PR #327, opened 2025-04-30
Improve memory efficiency

arjunshajitech:
Refactored packet processing to avoid multiple make calls and eliminate the use of append. This reduces garbage collection overhead and improves overall memory management efficiency.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/nack/generator_interceptor.go
+++ b/pkg/nack/generator_interceptor.go
@@ -145,6 +145,9 @@ func (n *GeneratorInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 
 	senderSSRC := rand.Uint32() // #nosec
 
+	missingPacketSeqNums := make([]uint16, n.size)
+	filteredMissingPacket := make([]uint16, n.size)
+
 	ticker := time.NewTicker(n.interval)
 	defer ticker.Stop()
 	for {
@@ -155,7 +158,7 @@ func (n *GeneratorInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 				defer n.receiveLogsMu.Unlock()
 
 				for ssrc, receiveLog := range n.receiveLogs {
-					missing := receiveLog.missingSeqNumbers(n.skipLastN)
+					missing := receiveLog.missingSeqNumbers(n.skipLastN, missingPacketSeqNums)
 
 					if len(missing) == 0 || n.nackCountLogs[ssrc] == nil {
 						n.nackCountLogs[ssrc] = map[uint16]uint16{}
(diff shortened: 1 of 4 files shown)


pkg/nack/receive_log.go
```diff
@@ -107,14 +107,15 @@ func (s *receiveLog) missingSeqNumbers(skipLastN uint16) []uint16 {
 		return nil
 	}
 
-	missingPacketSeqNums := make([]uint16, 0)
+	c := 0
 	for i := s.lastConsecutive + 1; i != until+1; i++ {
 		if !s.getReceived(i) {
-			missingPacketSeqNums = append(missingPacketSeqNums, i)
+			missingPacketSeqNums[c] = i
```
JoeTurki, line 113: I think we need protection against overflow here, if size < c?

pkg/nack/generator_interceptor.go
```diff
@@ -196,7 +205,7 @@ func (n *GeneratorInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 						}
 					}
 
-					if len(filteredMissing) == 0 {
+					if len(filteredMissingPacket) == 0 {
```
JoeTurki, line 208: This will never trigger now, since missingPacketSeqNums is always initialized with the interceptor size?

JoeTurki (changes requested):
The PR makes sense to me, but after looking at the code I found those two concerns.

Thank you so much.

pkg/nack/generator_interceptor.go
arjunshajitech: @JoeTurki Yes, you're right — that was my bad. I've pushed a fix, could you please reverify? Thank you.

pkg/nack/receive_log.go
arjunshajitech: I believe I’ve already added a test case for this scenario, and the code behaves as expected.

Assume the ``receiverLog size is 128.``

Initially, packet 101 is received. At this point, the lastConsecutive is 101.

Next, packet 250 is received. While adding the packet, Since 250 is beyond the window size of 128, it falls under the logic at:
https://github.com/pion/interceptor/blob/96a23bb5f5f80e084358d9392183e2ebb2076a9e/pkg/nack/receive_log.go#L71
https://github.com/pion/interceptor/blob/96a23bb5f5f80e084358d9392183e2ebb2076a9e/pkg/nack/receive_log.go#L72
These lines reset lastConsecutive, which is guaranteed to be less than 128.

Based on this, I don’t believe an overflow occurs.

Please feel free to correct me if I’m mistaken. Thank you.

JoeTurki (approved):
I've tested it manually, and it seems to re-produce all the old behavior exactly. The only trade-off is that we now allocate two fixed `n.size` slices each interval, which in a perfect conditions does introduce a small constant overhead? (It'll be nothing after all). please correct me if I'm wrong.

As a side note: we need tests for a few of the more corner-case path. There are a few more mutations more than I'm comfortable with (Not your PR, the behavior of the code itself before your PR) This is why I run manual tests :) I'll make future PRs for tests.

Thanks for the great work on this! This is great!

Issue #318, opened 2025-03-31
Flex FEC send from pion cause frame tears in browser.


3DRX:
### Your environment.

- github.com/pion/webrtc/v4 v4.0.14
- github.com/pion/interceptor v0.1.37
- Any browser with relative new version, desktop chromium and mobile safari in my case.

### What did you do?

I use flexfec03 via the following code

```go
	err = m.RegisterCodec(
		webrtc.RTPCodecParameters{
			RTPCodecCapability: webrtc.RTPCodecCapability{
				MimeType:     webrtc.MimeTypeFlexFEC + "-03",
				ClockRate:    9000,
				Channels:     0,
				SDPFmtpLine:  "repair-window=10000000",
				RTCPFeedback: nil,
			},
			PayloadType: 118,
		},
		webrtc.RTPCodecTypeVideo,
	)

// ...

	fecInterceptor, err := flexfec.NewFecInterceptor()
	if err != nil {
		panic(err)
	}
	i.Add(fecInterceptor)
```

Through logging in sender and observing receiver stats I'm pretty sure pion is sending encoded FEC packets correctly. But when I randomly drop some packets in a controlled network, the framerate in clientside didn't drop significantly, but the video tears apart from time to time. I think this perhaps means the browser can recover frames from FEC paload, but the recovered frame is incorrect.

3DRX:
This is the slow-downed version of the original 120fps video, for easier observation of the tearing.

https://github.com/user-attachments/assets/2736ca29-60e3-47c6-aed5-bca13971a35e

I'm pretty sure this is an issue related to FEC is because without FEC, with other parameters exactly the same, the video doesn't appears to be broken in it's content like this, it will only drop frames when facing lossy network.

aalekseevx:
@3DRX, it seems that the flexfec encoder is generating malformed packets. Can you please check if you have these issues on my branch at https://github.com/pion/intercepter/pull/330 ?

3DRX:
@aalekseevx Wow thanks you Aleksandr! I'll definitely check it out. I'll try to report back to you as soon as possible.

aalekseevx:
@3DRX, Hi, Is it hard to reproduce this issue? SDP might be helpful. A packet dump would help, unless it's an encrypted SRTP. Do you see a fecPacketsReceived plot in chrome://webrtc-internal stats? Maybe there's another issue with the interceptor not using the correct pt/ssrc. I'll try to reproduce it / build example code one I have some free time.

[...]

PR #330, opened 2025-05-18
Fix malformed FlexFEC-03

aalekseevx:
#### Description

Fix several issues with FlexFec-03 encoder, add simple decoder and round-trip tests

#### Reference issue
Fixes https://github.com/pion/interceptor/issues/318

### DIFF
--- a/pkg/flexfec/flexfec_03_test.go
+++ b/pkg/flexfec/flexfec_03_test.go
@@ -1,0 +1,119 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package flexfec
+
+import (
+	"testing"
+
+	"github.com/pion/rtp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	payloadType         = uint8(49)
+	ssrc                = uint32(867589674)
+	protectedStreamSSRC = uint32(476325762)
+)
+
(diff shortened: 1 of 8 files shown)


3DRX (approved):
A suggestion [here](https://github.com/pion/interceptor/issues/318#issuecomment-2892516820) to make this work out of the box for everyone. I've tested this PR with these additional changes, and this interceptor itself doesn't seems to have any problem. I also created a testing repo for this, [here](https://github.com/3DRX/fec-test).

aalekseevx:
@3DRX, I don’t want to make this PR even bigger and harder to review, so I created a separate issue regarding EncoderInterceptor with your suggestions and a couple of mine: https://github.com/pion/interceptor/issues/332. In this PR, I want to focus on fixing the encoder. I will make another one for the interceptor!

aalekseevx:
@JoeTurki, do you want to take a look at the code too? I'm not rushing you in any way, just want to make sure we're on the same page.

JoeTurki:
@aalekseevx 

The change looks good to me, great stuff, thank you so much.
The only nitpick is that I don't see a test for reallocation from variable-size frames. just to test the slice-growth fix under test. Because that part is easy to break in future changes.

3DRX:
> @3DRX, I don’t want to make this PR even bigger and harder to review, so I created a separate issue regarding EncoderInterceptor with your suggestions and a couple of mine: #332. In this PR, I want to focus on fixing the encoder. I will make another one for the interceptor!

Sure, thanks!

aalekseevx:
@JoeTurki, good catch! I have added the corresonding test

PR #349, opened 2025-07-19
Allow users to provide a custom logger implementation

mengelbart:
#### Description

The current implementation allows users to specify custom writers and custom formatters. But now Go has built-in structured logging, so I would like to do something like this:
```
slog.Info("received RTP packet", "sequence-number", pkt.header.SequenceNumber)
```

I think this is not possible with the current implementation. This PR allows me to provide a custom logger that can then implement any form of custom logging.


JoeTurki (approved):
lgtm

PR #351, opened 2025-07-31
Rename and deprecated rfc8888 package

mengelbart:
Rename `rfc8888` packages. We don't have any other packages named after the respective RFCs.


JoTurk (approved):

JoTurk:
I think we can ignore the API check this time, since interceptor is v0 and I think this change is compatible enough.

mengelbart:
Sounds good, I'll merge it now. Sorry, I didn't merge it earlier. I kind of lost track of this.

PR #354, opened 2025-08-10
Remove unused named return parameters

cppcoffee:
Remove unused named return parameters from `FindNextAtOrAfter` method as they were not being used in the function implementation.

### DIFF
--- a/pkg/twcc/arrival_time_map.go
+++ b/pkg/twcc/arrival_time_map.go
@@ -104,13 +104,13 @@ func (m *packetArrivalTimeMap) EndSequenceNumber() int64 {
 }
 
 // FindNextAtOrAfter returns the sequence number and timestamp of the first received packet that has a sequence number
-// greator or equal to sequenceNumber.
+// greater or equal to sequenceNumber.
 func (m *packetArrivalTimeMap) FindNextAtOrAfter(sequenceNumber int64) (
-	foundSequenceNumber int64, arrivalTime int64, ok bool,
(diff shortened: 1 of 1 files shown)


pkg/twcc/arrival_time_map.go
```diff
@@ -106,11 +106,11 @@ func (m *packetArrivalTimeMap) EndSequenceNumber() int64 {
 // FindNextAtOrAfter returns the sequence number and timestamp of the first received packet that has a sequence number
 // greator or equal to sequenceNumber.
```
JoeTurki, line 107: Not related to your PR but there is a typo `greator`.

pkg/twcc/arrival_time_map.go
```diff
@@ -106,11 +106,11 @@ func (m *packetArrivalTimeMap) EndSequenceNumber() int64 {
...
 func (m *packetArrivalTimeMap) FindNextAtOrAfter(sequenceNumber int64) (
-	foundSequenceNumber int64, arrivalTime int64, ok bool,
```
JoeTurki, line 109: I like this change but I think it makes this function slightly less readable from `foundSequenceNumber int64, arrivalTime int64, ok bool` int64, int64, bool, maybe we can move this to the function comment? or name the local variables to foundSequenceNumber and arrivalTime instead of seq and t.

JoeTurki (commented):
Thank you for all the improvements you do <3

PR #365, opened 2025-09-09
Resulting report from BuildReport() was exceeding the provided maxSize limit

astroza:
#### Description

Under packet loss stress, the rfc8888 interceptor was sending packets bigger than the maxSize limit, sometimes even bigger than the network MTU.

I found the next problems:

1. Ranging over a map is not deterministic (this applies only for the testing code)
    ```	
    for i, log := range r.streams {
        if len(r.streams) > 1 && int(i) == len(r.streams)-1 {
            maxReportBlocksPerStream = maxReportBlocks % len(r.streams)
        }
        block := log.metricsAfter(now, int64(maxReportBlocksPerStream))
        report.ReportBlocks = append(report.ReportBlocks, block)
   }
    ```
    maxReportBlocksPerStream could be updated in the last iteration as well as the first one
2. r.streams is a "ssrc -> log" map, so the `i` variable will never be  equal to len(r.streams)-1. `i` is not a array index

3. maxReportBlocksPerStream as maxReportBlocks / (len(r.streams)-1) will exceed the blocks limit. This is specially true when len(r.streams) is a small number

4. The tests were not catching the actual issue
     If you run the new test "MaxreportsPerStream 3 streams" with the old code, you will get 
     ```
             	Error Trace:	/home/fastroza/src/interceptor/pkg/rfc8888/recorder_test.go:166
        	Error:      	"1746" is not less than "1200"
      ```


Sean-Der:
Nice catch :) 

@mengelbart what you think? If I don't hear back in a day or two @astroza I will merge myself!

nils-ohlmeier (approved):

PR #374, opened 2025-10-15
Add interceptor to aggregate CCFB reports

mengelbart:
Alternative but cleaner implementation for congestion control feedback receiver (#300). The main difference between this and #300 is that #300 organizes the feedback per stream, while this version organizes it as a single linear stream, which is easier to process. If one still needs to know which stream a packet was sent on, that information is still included in the reports for each packet.

### DIFF
--- a/pkg/rtpfb/interceptor.go
+++ b/pkg/rtpfb/interceptor.go
@@ -39,6 +39,14 @@ type packetLog interface {
 // Option can be used to set initial options on CCFB interceptors.
 type Option func(*Interceptor) error
 
+func WithLoggerFactory(lf logging.LoggerFactory) Option {
+	return func(i *Interceptor) error {
+		i.logFactory = lf
+
+		return nil
+	}
+}
+
 func timeFactory(f func() time.Time) Option {
 	return func(i *Interceptor) error {
 		i.timestamp = f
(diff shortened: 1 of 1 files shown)


pkg/rtpfb/ccfb_receiver.go
```diff
@@ -0,0 +1,70 @@
...
+	for i, mb := range blocks {
+		if mb.Received {
+			arrival := time.Time{}
+
+			// RFC 8888 states: If the measurement is unavailable or if the
+			// arrival time of the RTP packet is after the time represented by
+			// the RTS field, then an ATO value of 0x1FFF MUST be reported for
+			// the packet. In that case, we set a zero time.Time value.
+			if mb.ArrivalTimeOffset != 0x1FFF {
+				delta := time.Duration((float64(mb.ArrivalTimeOffset) / 1024.0) * float64(time.Second))
```
JoeTurki: maybe we should do this to avoid drift from using float?
```suggestion
				delta := time.Duration(mb.ArrivalTimeOffset) * time.Second / 1024
```

JoeTurki (commented):
Just two fast comments / nits, I'll read the RFC again and will actually review :)

pkg/rtpfb/interceptor.go
```diff
@@ -0,0 +1,238 @@
...
+	return &InterceptorFactory{
+		opts: opts,
+	}, nil
+}
+
+// NewInterceptor returns a new ccfb.Interceptor.
+func (f *InterceptorFactory) NewInterceptor(_ string) (interceptor.Interceptor, error) {
+	in := &Interceptor{
+		NoOp:      interceptor.NoOp{},
+		log:       logging.NewDefaultLoggerFactory().NewLogger("ccfb_interceptor"),
```
sirzooro: Please add support for custom `LoggerFactory` as in https://github.com/pion/interceptor/pull/396

PR #381, opened 2025-11-13
ECN support

penhauer:
Just updated `ecn`'s type from  `uint8` to `rtcp.ECN` plus actually reading the ECN value from the interceptor attributes.


mengelbart (changes requested):

JoTurk:
I rebased and simplifed this PR.

JoTurk (approved):

pkg/ccfb/interceptor.go
```diff
@@ -122,11 +122,12 @@ func (s *SenderInterceptor) BindRemoteStream(
 			return 0, nil, err
 		}
 
+		ecn, _ := attr.Get(interceptor.ECNKey).(rtcp.ECN)
 		p := packet{
 			arrival:        s.now(),
 			ssrc:           header.SSRC,
 			sequenceNumber: header.SequenceNumber,
-			ecn:            0, // ECN is not supported (yet).
+			ecn:            uint8(ecn),
```
mengelbart: we could convert the ecn field to rtcp.ECN, too, because that is what it will be converted back to when we build the report.

pkg/ccfb/interceptor_test.go
```diff
@@ -348,3 +348,70 @@ func TestConcurrentClose(t *testing.T) {
...
+		assert.NoError(t, marshalErr)
+		var report rtcp.CCFeedbackReport
+		assert.NoError(t, report.Unmarshal(raw))
+		assert.Len(t, report.ReportBlocks, 1)
+		block := report.ReportBlocks[0]
+		assert.Equal(t, header.SSRC, block.MediaSSRC)
+		assert.Zero(t, block.BeginSequence)
+		assert.Len(t, block.MetricBlocks, 2*len(cases)-1)
+		for i, metric := range block.MetricBlocks {
+			assert.Equal(t, i%2 == 0, metric.Received)
```
mengelbart: I don't get this testcase. Why do we assert marshal/unmarshal here and why the i%2 and sequence number skipping? Seems like it is testing more than what's needed to assert ECN handling.

pkg/ccfb/interceptor.go
JoTurk: It's unfortunate because the public API uses uint8 `Recorder.AddPacket(..., ecn uint8)` I can store it as rtcp.ECN but we'll have a mismatch in types until we break the public API.

pkg/ccfb/interceptor_test.go
JoTurk: Sorry the test was from when the PR was more complex and I didn't trim it.

JoTurk: Trimmed.

pkg/ccfb/interceptor.go
JoTurk: I'll do that and move the cast to AddPacket but the other side. Until we can break `Recorder.AddPacket`

JoTurk: Done we now cast 3 times instead of 1 before :) but we can get rid of them once we break the public API :)

mengelbart: Oh, sorry, I didn't realize that API is public. I'm not sure if there are any downsides of having rtcp.ECN in the public API of this package. If not, I'm fine with breaking it at any time since we're still v0. (I'd also be fine with unexporting the recorder, but I guess that may be really annoying if anyone is using it already...).

mengelbart (approved):

pkg/ccfb/interceptor.go
JoTurk: searched both package names `"github.com/pion/interceptor/pkg/rfc8888" "NewRecorder(" language:Go -repo:pion/interceptor` and it doesn't seem that anyone is using this API, I'll break it \:)

mengelbart: It might break the deprecated RFC8888 package that just redirects here, but that should be fine, too.

JoTurk: yeah it broke both rfc8888 and ccfb, but removed the casts, from a quick github search nobody used this API from the old package or the new.

mengelbart: We could just drop rfc8888 already, but we can also do it separately from this PR.

Issue #328, opened 2025-05-09
Possible memory leak on nack package


tubzby:
I have an SFU program using ion-sfu (modified a bit since the original repo is outdated), which depends on interceptor v0.1.37.

Recently, we noticed that the memory usage of this process has been increasing over the past few days, the graph looks like:

![Image](https://github.com/user-attachments/assets/85be8abf-61d5-4d19-b9e8-744ef0245ffb)

So I use `go tool pprof` to get the heap output and then found out that most of the active memory was allocated by the nack package:

![Image](https://github.com/user-attachments/assets/69fed704-d006-4db0-8e82-c21cf1f007e0)

![Image](https://github.com/user-attachments/assets/39d4e6cd-b922-4ee8-bcd6-f485ed21c61c)

top of pprof:

```console
(pprof) top
Showing nodes accounting for 4.89GB, 88.17% of 5.55GB total
Dropped 522 nodes (cum <= 0.03GB)
Showing top 10 nodes out of 106
      flat  flat%   sum%        cum   cum%
    3.55GB 63.91% 63.91%     3.55GB 63.91%  github.com/pion/interceptor/pkg/nack.newPacketManager.func2
    0.41GB  7.38% 71.30%     0.41GB  7.38%  github.com/pion/webrtc/v4/pkg/media/samplebuilder.New
    0.28GB  5.04% 76.34%     0.28GB  5.04%  gdcx.com/sfu/cluster.NewSFU.NewBufferFactory.func1
    0.24GB  4.25% 80.59%     3.97GB 71.49%  github.com/pion/interceptor/pkg/nack.(*packetManager).NewPacket
    0.18GB  3.33% 83.91%     0.18GB  3.33%  github.com/pion/interceptor/pkg/nack.newPacketManager.func1
    0.09GB  1.54% 85.45%     0.09GB  1.54%  github.com/pion/transport/v3/packetio.(*Buffer).grow
    0.05GB  0.87% 86.32%     0.05GB  0.87%  syscall.NetlinkRIB
    0.04GB  0.69% 87.00%     0.04GB  0.69%  log.New
    0.03GB  0.61% 87.61%     0.03GB  0.61%  github.com/pion/sdp/v3.(*unmarshalCache).cloneMediaAttributes
    0.03GB  0.56% 88.17%     0.05GB  0.81%  github.com/pion/webrtc/v4/pkg/media/samplebuilder.(*SampleBuilder).buildSample
(pprof)
```
[...]

PR #386, opened 2025-12-01
Fix memory leak with nackCountLogs

philipch07:
#### Description
`nackCountLogs` wasn't getting cleaned up properly in `generator_interceptor.go` from the nack package.

this is fixed in two ways:
1) `UnbindRemoteStream()` can delete the respective `nackCountLogs` for the provided `SSRC`.
2) in the `loop()` we can drop the `nackCountLogs` map if it's empty. This is particularly helpful if a lot of them get accumulated over time.

#### Reference issue
#328 and @JoeTurki mentioned that there were reports of a memory leak so hopefully this fixes it.


JoeTurki (approved):
Yeah, as I promised to test it again when i wake up it's working as intended 😄

PR #387, opened 2025-12-01
Fix nack deadlocks from writers

### LINKED ISSUE #384: Interceptor.loop() goroutine leak
### Your environment.
  - Version: v.0.1.37
  - Browser: Firefox - 129.0.2, Chrome - 138, Safari - not tested

### What did you do?
Hello everyone, this is my first GitHub issue, so please guide me if I am doing something wrong. I am trying to create a simple server which will accept requests from a browser and create a WebRTC connection with audio track. I noticed that if I reload the page shortly after connection is established, goroutine leak will be created. 
Here are results from profiler:
Image

**Important:** I can reproduce this issue only when the server is located on a remote host. I cannot reproduce it on the local server.

### What did you expect?
WebRTC connection to close properly.

### What happened?
PeerConnection is closing partially, and the interceptor goroutines are never closing.
Seams like PeerConnection.close() stuck during execution. To be more specific it stuck at iceTransport.GracefulStop():
repository - **webrtc**, file - **peerconnection.go:**
`2443    if pc.iceTransport != nil {`
`2444    	here -->		gracefulCloseErrors = append(gracefulCloseErrors, pc.iceTransport.GracefulStop())`
`2445    	}`

I investigated it a little bit, and i think problem is in handlerNotifier.Close() function. If Close() is graceful we will need to wait for WaitGroup to unblock the execution:
repository - **ice**, file - **agent_handlers.go:**
`68	if graceful {`
`69		// if we were closed ungracefully before, we now`
`70		// want ot wait.`
`71 here -->  defer h.notifiers.Wait()`
`72	} `
I assume that  handlerNotifier.EnqueueConnectionState() function never gets to "WaitGroup.done()" for some reason.

philipch07:
#### Description
There were two deadlocks that I found from the generator and responder interceptors.

Each of these had issues from writing while holding a mutex, which could cause deadlocks. I wrote a test that hits the deadlock in the old implementation (and fails) but passes with the fix.

Note that this branch is based off of #386, so ideally that should get merged first.

#### Reference issue
Fixes #384

### DIFF
--- a/pkg/nack/generator_interceptor.go
+++ b/pkg/nack/generator_interceptor.go
@@ -156,66 +156,71 @@ func (n *GeneratorInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 	for {
 		select {
 		case <-ticker.C:
-			func() {
-				n.receiveLogsMu.Lock()
-				defer n.receiveLogsMu.Unlock()
+			// save NACKs to send without holding the mutex during Write
+			var toSend []rtcp.Packet
 
-				for ssrc, receiveLog := range n.receiveLogs {
-					missing := receiveLog.missingSeqNumbers(n.skipLastN, missingPacketSeqNums)
+			n.receiveLogsMu.Lock()
+			for ssrc, receiveLog := range n.receiveLogs {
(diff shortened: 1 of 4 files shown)


Sean-Der (approved):
love the tests :)

PR #390, opened 2025-12-17
Avoid RTPBuffer allocation

arjunshajitech:
#### Description

Previously, NewInterceptor validated the RTP buffer size by calling rtpbuffer.NewRTPBuffer
this resulted in unnecessary memory allocations during interceptor construction, even though
the buffer itself was not used at that stage.

#### Reference issue
Fixes #...

### DIFF
--- a/pkg/nack/receive_log.go
+++ b/pkg/nack/receive_log.go
@@ -4,7 +4,6 @@
 package nack
 
 import (
-	"fmt"
 	"sync"
 
 	"github.com/pion/interceptor/internal/rtpbuffer"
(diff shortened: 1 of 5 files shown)


JoTurk (commented):
cool change, thank you.

pkg/nack/receive_log.go
```diff
@@ -20,19 +19,8 @@ type receiveLog struct {
...
-			correctSize = true
-
-			break
-		}
-		allowedSizes = append(allowedSizes, 1<<i)
-	}
-
-	if !correctSize {
-		return nil, fmt.Errorf("%w: %d is not a valid size, allowed sizes: %v", ErrInvalidSize, size, allowedSizes)
+	if err := rtpbuffer.IsBufferSizeValid(size); err != nil {
```
mengelbart, line 22: This now uses `rtpbuffer.IsBufferSizeValid` to check if we can create a `receiveLog`. What if we ever change the conditions for only one of them? Could we rename the `IsBufferSizeValid` function to `IsPowerOfTwo` and move it out of the `rtpbuffer` package? That way, it would be clear what it does, and if we change the conditions of either `receiveLog` or `RTPBuffer`, it would not affect the other.

arjunshajitech: @mengelbart I’ve updated it to use a separate `IsPowerOfTwo` helper instead of `rtpbuffer.IsBufferSizeValid`

mengelbart (approved):

PR #396, opened 2026-01-08
Added support for custom LoggerFactory

sirzooro:
Added options to interceptors to pass LoggerFactory instance and use it to create loggers for interceptors.


JoTurk:
Thank you. I think we should name the option WithLoggerFactory to stay consistent with the ICE. also with SCTP, and DTLS options (which are currently in progress). What do you think?

PR #414, opened 2026-04-24
cc: trim OnTransportCCFeedback result to PacketStatusCount

AnshulMalik:
StatusVectorChunks always encode a fixed number of slots (7 for 2-bit symbols, 14 for 1-bit). When the number of reported packets does not fill the final chunk, the encoder pads the remaining slots with TypeTCCPacketNotReceived symbols.

Without trimming, those padding slots are looked up in the send history and, if found (e.g. recently sent packets whose SNs happen to fall in the padding range), are returned as Acknowledgments with a zero Arrival time. The loss estimator then counts them as lost packets, inflating the loss estimate even though those packets are still in flight.

### DIFF
--- a/internal/cc/feedback_adapter.go
+++ b/internal/cc/feedback_adapter.go
@@ -182,6 +182,11 @@ func (f *FeedbackAdapter) OnTransportCCFeedback(
 		}
 	}
 
+	// Discard padding slots from the last StatusVectorChunk.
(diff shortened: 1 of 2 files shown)


mengelbart:
LGTM. Thanks for the PR @AnshulMalik! Would you mind fixing the commit message? I think you just need to remove the `cc: ` part and capitalize the first word.

PR #417, opened 2026-06-01
Read reception reports from sender report packets

### LINKED ISSUE #412: Stats interceptor ignores ReportBlocks embedded in incoming SenderReports
### Your environment.
  - Version: 4.0.0
  - Browser: Chromium 139.0.7258.5 (build 1181) 

We are acting as a softphone endpoint to a number of CCaaS vendor platforms, and one problem we've been having is successfully extracting the remote RTCP statistics from the connection using the standard interceptor  in the case of one of the vendors...

pkg/stats/stats_recorder.go → recordIncomingRTCP() does not process the Reports (ReportBlocks) inside an incoming *rtcp.SenderReport. Only standalone *rtcp.ReceiverReport packets populate RemoteInboundRTPStreamStats (jitter, RTT, fraction lost, packets lost). When a remote peer embeds its RR report blocks inside an SR — which is perfectly valid per RFC 3550 §6.4.1 — those stats are silently discarded.

We added rtcp.Unmarshal() parsing to our RTCP drain goroutine to inspect the raw packets arriving from each vendor. NICE sends a compound RTCP packet every ~6 seconds with this structure:

Packet 1 (80 bytes): SenderReport + SDES
  SenderReport ssrc=610093063 packetCount=257 reportBlocks=1
    ReportBlock[0] ssrc=3507758364 jitter=25 fractionLost=0 totalLost=0
                   lastSR=2093454453 delaySinceLastSR=50266
                   (matches our OutboundSSRC ✓)

Packet 2 (44 bytes): ExtendedReport (XR)
The ReportBlock data is correct and real — SSRC matches our outbound track, jitter is non-zero, lastSR and delaySinceLastSR are populated (sufficient for RTT calculation). But Pion's stats interceptor returns all zeros for RemoteInboundRTPStreamStats.
By contrast, vendors that produce working stats (Genesys, Amazon Connect) send standalone ReceiverReport packets, which the interceptor handles correctly.

Bit more info, not sure if I've got this right... In stats_recorder.go, recordIncomingRTCP() line 291:

case *rtcp.SenderReport:
    latestStats.RemoteOutboundRTPStreamStats.PacketsSent = uint64(pkt.PacketCount)
    latestStats.RemoteOutboundRTPStreamStats.BytesSent = uint64(pkt.OctetCount)
    latestStats.RemoteTimeStamp = ntp.ToTime(pkt.NTPTime)
    latestStats.ReportsSent++
It extracts the SR header fields but never iterates pkt.Reports — the embedded ReportBlocks that contain the same jitter/loss/RTT data as a standalone ReceiverReport.
The recordIncomingRR() method already handles this data correctly, but it's only called for standalone *rtcp.ReceiverReport packets (line 289-290).
Suggested fix
Process the SR's embedded report blocks by calling recordIncomingRR with a synthesized ReceiverReport:

case *rtcp.SenderReport:
    latestStats.RemoteOutboundRTPStreamStats.PacketsSent = uint64(pkt.PacketCount)
    latestStats.RemoteOutboundRTPStreamStats.BytesSent = uint64(pkt.OctetCount)
    latestStats.RemoteTimeStamp = ntp.ToTime(pkt.NTPTime)
    latestStats.ReportsSent++

    // Process embedded report blocks (RR data inside SR)
    if len(rtcpPkt.Reports) > 0 {
        rr := &rtcp.ReceiverReport{SSRC: rtcpPkt.SSRC, Reports: rtcpPkt.Reports}
        latestStats = r.recordIncomingRR(latestStats, rr, v.ts)
    }
This is a two-line addition. recordIncomingRR already does the SSRC matching, jitter/loss extraction, and RTT computation (matching LastSenderReport against stored outgoing SR NTP timestamps).

for the moment our workaround which works well without the upstream change is, we parse RTCP ourselves in our sender.Read() drain goroutine. When we see an incoming SenderReport with report blocks matching our outbound SSRC, we manually extract jitter, fraction lost, total lost, and compute RTT by reconstructing the full NTP timestamp from the compact 32-bit LastSenderReport value. We store these on the session and use them as a fallback when RemoteInboundRTPStreamStats returns zeros.

mengelbart:
#### Description
Fix for reading reception reports blocks from Sender Reports.

#### Reference issue
Fixes #412

### DIFF
--- a/pkg/stats/stats_recorder.go
+++ b/pkg/stats/stats_recorder.go
@@ -236,8 +236,12 @@ func (r *recorder) recordOutgoingRTP(latestStats internalStats, v *outgoingRTP) 
 	return latestStats
 }
 
-func (r *recorder) recordIncomingRR(latestStats internalStats, pkt *rtcp.ReceiverReport, ts time.Time) internalStats {
-	for _, report := range pkt.Reports {
+func (r *recorder) recordIncomingRR(
+	latestStats internalStats,
+	reports []rtcp.ReceptionReport,
+	ts time.Time,
+) internalStats {
+	for _, report := range reports {
 		if latestStats.remoteInboundFirstSequenceNumberInitialized {
 			cycles := uint64(report.LastSequenceNumber&0xFFFF0000) >> 16
 			nr := uint64(report.LastSequenceNumber & 0x0000FFFF)
@@ -312,12 +316,13 @@ func (r *recorder) recordIncomingRTCP(latestStats internalStats, incoming *incom
 		case *rtcp.PictureLossIndication:
 			latestStats.OutboundRTPStreamStats.PLICount++
 		case *rtcp.ReceiverReport:
-			latestStats = r.recordIncomingRR(latestStats, pkt, incoming.ts)
+			latestStats = r.recordIncomingRR(latestStats, pkt.Reports, incoming.ts)
 		case *rtcp.SenderReport:
 			latestStats.RemoteOutboundRTPStreamStats.PacketsSent = uint64(pkt.PacketCount)
 			latestStats.RemoteOutboundRTPStreamStats.BytesSent = uint64(pkt.OctetCount)
 			latestStats.RemoteTimeStamp = ntp.ToTime(pkt.NTPTime)
 			latestStats.ReportsSent++
+			latestStats = r.recordIncomingRR(latestStats, pkt.Reports, incoming.ts)
 
 		case *rtcp.ExtendedReport:
 			return r.recordIncomingXR(latestStats, pkt, incoming.ts)
--- a/pkg/stats/stats_recorder_test.go
+++ b/pkg/stats/stats_recorder_test.go
@@ -927,6 +927,68 @@ func TestStatsRecorder(t *testing.T) {
 				},
 			},
 		},
+		{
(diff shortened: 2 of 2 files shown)


pkg/stats/stats_recorder.go
```diff
@@ -236,8 +236,12 @@ func (r *recorder) recordOutgoingRTP(latestStats internalStats, v *outgoingRTP)
...
 }
 
-func (r *recorder) recordIncomingRR(latestStats internalStats, pkt *rtcp.ReceiverReport, ts time.Time) internalStats {
-	for _, report := range pkt.Reports {
+func (r *recorder) recordIncomingRR(
+	latestStats internalStats,
+	reports []rtcp.ReceptionReport,
+	ts time.Time,
+) internalStats {
+	for _, report := range reports {
```
JoTurk, line 244: question should we check if the report.ssrc matches the recorder ssrc?

JoTurk (approved):
looks good thank you.

pkg/stats/stats_recorder.go
mengelbart: Good point, we probably should. I think this is also relevant for the other RTCP packet types, though. I will create a separate issue.

PR #426, opened 2026-07-14
Reduce TWCC sender allocations

kcaffrey:
`Recorder.Record` heap-allocates on every call: it stores the unwrapped sequence number into `startSequenceNumber` by taking the address of a local, which moves the local to the heap whether or not the branch is taken. Since `Record` runs once per received RTP packet, this costs one allocation per packet. Updates now go through a setter that writes into a single reused allocation.

`feedback.addReceived` also allocated one `*rtcp.RecvDelta` per received packet. Deltas are now accumulated by value, and the `[]*rtcp.RecvDelta` that the rtcp API requires is materialized once per feedback packet in `getRTCP`, which hands ownership of the backing array to the packet.

No exported API or behavior changes.

-----------

Benchmarks (added in this PR, run on main vs this branch on a M2 MacBook Air; the second records 100 packets per feedback):

|  | main | this PR |
|---|---|---|
| `BenchmarkRecord` | 21.1 ns/op, 1 alloc/op | 6.9 ns/op, 0 allocs/op |
| `BenchmarkRecordAndBuildFeedbackPacket` | 4546 ns/op, 322 allocs/op | 2270 ns/op, 21 allocs/op |

`TestRecordDoesNotAllocateInSteadyState` pins the steady-state `Record` path at zero allocations.

### DIFF
--- a/pkg/twcc/twcc.go
+++ b/pkg/twcc/twcc.go
@@ -52,7 +52,7 @@ func (r *Recorder) Record(mediaSSRC uint32, sequenceNumber uint16, arrivalTime i
 	unwrappedSN := r.sequenceUnwrapper.Unwrap(sequenceNumber)
 	r.maybeCullOldPackets(unwrappedSN, arrivalTime)
 	if r.startSequenceNumber == nil || unwrappedSN < *r.startSequenceNumber {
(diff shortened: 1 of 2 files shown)


JoTurk (approved):
thank you

Issue #278, opened 2024-10-07
Priority Queue has an infinite Loop bug


hafus:
### Your environment.
  - Version: *v0.1.36*
  - Browser: *no browser*

### What did you do?
The code snippet below is a test case that catches the reported bug.
```Go
t.Run("Infinite loop", func(*testing.T) {
		q := NewQueue()
		pkt := &rtp.Packet{Header: rtp.Header{SequenceNumber: 5000, Timestamp: 500}, Payload: []byte{0x02}}
		q.Push(pkt, pkt.SequenceNumber)
		q.Push(pkt, pkt.SequenceNumber)
		assert.Equal(uint16(1), q.Length())
		popped, _ := q.PopAt(uint16(5012))
		assert.Equal(popped.SequenceNumber, uint16(5000))
		assert.Equal(uint16(0), q.Length())

	})
```

### What did you expect?
Priority queue should have a length of one after inserting a duplicate packet of the first packet in the queue or at least not enter into an infinite loop.

### What happened?
The Push func runs for infinite time because the duplicated packet has head and prev pointers pointing into the head packet, resulting in a loop.
https://github.com/pion/interceptor/blob/0eab1887cec79081e21457c0f9f89be2537c5df5/pkg/jitterbuffer/priority_queue.go#L80-L100

### suggestions
I suggest dropping any duplicated packet because it already exists in the queue.

ahmedBahossain:
@Sean-Der, please have a look at the issue above.

hafus:
@Sean-Der, please have a look at the issue above.

Sean-Der:
Hey @hafus I would really appreciate your help on this.

I want to drop the Priority Queue and instead have the JitterBuffer and NACK Responder powered by the same data structure. Would you be interested in helping me with this?

hafus:
Hello @Sean-Der 

Of course, that's my pleasure. I will start working on it. If you have any suggestions or known issues about the current implementation, please share them with me.

PR #428, opened 2026-07-20
Use rtpbuffer for the jitter buffer

mojl:
#### Description
Use rtpbuffer for the jitter buffer (as suggested by Sean in #278). I am using a ring buffer (as rtpbuffer packets can be overwritten any time) for the playback buffer and it accepts retainable packets too so it uses the buffer pool too.

I tried to keep the behavior of the API as 1:1 as possible, but the differences are:
- Emitting starts when we have X amount of sequential packets, not just any packets (`minStartCount`).
- All forms of `Pop` do move the `playoutHead` to that sequance number.
- The default capacity is now 1024, as keyframes can be big sometimes.

Working on: receiver interceptor, unit tests for the ring buffer, and a NoCopy option.


mojl:
The interceptor now supports this new jitter buffer, supports multiple tracks/streams, and accepts options for building the jitter buffer.

I've added the #255 fix since it was stopping me from testing.

thatsnotright:
Let me do some performance regression tests but this looks good

thatsnotright:
I have a branch that changes the data structure underneath for jb from a priority queue to a tree, based on feedback from @at-wat with my original port of SampleBuilder to jitterbuffer. This PR would allocate a lot more (~24% more in the perf tests) and is slower by ~68%. I'll do some profiling to see where the issues may be and hopefully we can reduce the memory allocations.

mojl:
Thank you for checking this PR, Rob.

The data structure I've used (ring buffer) is dictated by the data structure the internal `rtpbuffer` already uses (circular ring?), I've used the simplest data structure possible that can work alongside it. It does outperform a priority queue a bit, but not by much and a priority queue is simpler and more predictable and I only used this to reuse code.

The performance degradation is due to using `RetainablePacket`, since that's what `rtpbuffer` accepts. With some memory reuse `ReceiverInterceptor` can outperform the current implementation by removing 1 allocation, this is possible now since we are using `PacketFactory` which copies on push per default.

But as I understand `ReceiverInterceptor` is pretty much unusable here since the user/dev cannot skip gaps using it, so they should be told to reuse memory when using `JitterBuffer` directly in documentation.

There is trade off, and this is the result of using `rtpbuffer`, the conclusion can be either that we want that trade off: faster with copy + memory reuse, or that we prefer to not use `rtpbuffer` here.

I am happy to keep working on this PR if you think it is worth it after the optimization, I will push a commit later.

PR #436, opened 2026-08-11
Avoid stalling when packet sizes exceed burst size

mengelbart:
Automatically increases MTU size, which is then used as the minimum burst.

### DIFF
--- a/pkg/pacing/interceptor.go
+++ b/pkg/pacing/interceptor.go
@@ -35,7 +35,7 @@ type Option func(*Interceptor) error
 // the interceptor factory.
 func InitialRate(rate int) Option {
 	return func(i *Interceptor) error {
-		i.initialRate = rate
+		i.rate = rate
 
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


fallenmi (approved):
Disclosure: I used Codex to assist with this review and verified the result against exact commit `86cb3fdaa60a361ddf73951b4ca832d3a9c753a2`.

I exercised the pacer at 1,499, 1,500, 1,501, 4,000, and 65,535-byte RTP packet sizes against the exact base, this head, and the current GitHub merge. The base sends 1,499 bytes but stalls at the fixed-burst boundary and above; this change sends every case, including exact equality, while preserving the largest observed packet size across later rate changes. A 100-repetition race-enabled concurrent rate/MTU update oracle, the focused pacing suite, vet, and the remaining full race suite passed. I found no blocking issues.

pkg/pacing/interceptor.go
```diff
@@ -174,18 +179,34 @@ type Interceptor struct {
...
+
+func (i *Interceptor) growMTU(bytes int) {
+	i.rateLock.Lock()
+	defer i.rateLock.Unlock()
+
+	if bytes <= i.mtu {
+		return
+	}
+	i.mtu = bytes
+	i.limit.SetRate(i.rate, burst(i.rate, i.interval, 8*bytes))
```
JoTurk, line 209: Are there any plans to restore / recompute MTU?

JoTurk (approved):

pkg/pacing/interceptor.go
mengelbart: I thought about it, but wasn't sure what a good strategy would be.

PR #438, opened 2026-08-17
GCC: Keep increase path monotonic

uartnet:
The steady-state increase branch caps its result at 1.5 times the received rate. When that cap is below the current target, the increase operation can unexpectedly reduce the target even though the controller remains in the increase state. The other increase branch already guards this invariant.

Keep the target unchanged when the computed rate would be lower, and cover the regression with a focused test.

Validated with go test ./... and go test -race ./pkg/gcc.


Awuqing (commented):
That's a great idea to improve "R < 2/3 T"

PR #461, opened 2026-09-14
Do not send Sender Reports before any RTP has been sent

tmatth:
The sender report interceptor emitted an SR for every bound stream on each tick, including streams that had not written a single packet. For those the RTP timestamp was derived from a zero `time.Time`, so the report mapped "now" to a garbage timestamp (2269117121 at 90 kHz, 923864645 at 48 kHz). `RFC 3550 6.4` says a source that has sent nothing sends an RR, which the receiver interceptor already does.
    
Receivers that anchor RTCP-based timing on the first SR they see (for example `OvenMediaEngine` with `RtcpBasedTimestamp`) then have their anchor jump when the first real SR arrives a second later, and the stream can be stuck with a PTS that moved backwards by hours.
    
Skip streams with no packets sent when generating reports. The two existing subtests that relied on an SR before any RTP are adjusted: "before any packet" now asserts nothing is written, and "inject ticker" writes one RTP packet before ticking.

This PR also adds a test specifically for this transition.


pkg/report/sender_interceptor.go
```diff
@@ -126,8 +126,10 @@ func (s *SenderInterceptor) loop(rtcpWriter interceptor.RTCPWriter) {
 			s.streams.Range(func(_, value any) bool {
 				if stream, ok := value.(*senderStream); !ok {
 					s.log.Warnf("failed to cast SenderInterceptor stream")
+				} else if report := stream.generateReport(now); report == nil {
```
mengelbart, line 129: Could we break up the if/else branching into multiple `if ... {return true}` steps to avoid an empty branch? I find the flow a bit confusing here.

pkg/report/sender_interceptor_test.go
```diff
@@ -247,10 +245,107 @@ func TestSenderInterceptor(t *testing.T) {
...
+		assert.Equal(t, &rtcp.SenderReport{
+			SSRC:        123456,
+			NTPTime:     ntp.ToNTP(mNow.Now()),
+			RTPTime:     90000,
+			PacketCount: 1,
+			OctetCount:  3,
+		}, sr)
+	})
+
+	t.Run("write error is logged and loop keeps running", func(t *testing.T) {
```
mengelbart, line 309: I think this doesn't actually assert the logging, so maybe remove that from the test case name?

mengelbart (commented):
Thanks, I think this makes sense.

JoTurk (approved):
Looks good, I agree with @mengelbart comments

mengelbart (approved):

